PR #2, opened 2016-08-15
Use monotonic time

harlowja:
Monotonic time has the nice benefit of never
going backwards or getting messed up by NTPd
so use it instead of time.time for computing
durations (its a built-in on py3.x so we might
as well).

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,3 @@
 six>=1.7.0
 futures>=3.0;python_version=='2.7' or python_version=='2.6' # BSD
+monotonic>=0.6 # Apache-2.0
--- a/tenacity.py
+++ b/tenacity.py
@@ -15,11 +15,12 @@
 # limitations under the License.
 
 import random
-import six
 import sys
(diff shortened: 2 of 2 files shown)


requirements.txt
```diff
@@ -1 +1,2 @@
 six>=1.7.0
+monotonic>=0.6 # Apache-2.0
```
jd: Should we have a requirement only on Python < 3.something?

harlowja: I can, although how this package is coded it imports the py3 version and just uses that when it can.

https://github.com/atdt/monotonic/blob/master/monotonic.py#L51

We can do the same thing if we want, or just let the module there  do it.

jd: Yeah no, that would be silly, good catch!

jd:
@harlowja if you can resolve the conflicts… :D

harlowja:
I could :-P

harlowja:
Conflicts resolved.

Issue #4, opened 2016-08-15
3.2 and 3.3 support?


harlowja:
Why have these?

```
Programming Language :: Python :: 3.2

Programming Language :: Python :: 3.3
```
Remove them?

jd:
I don't mind. The code currently works on 3.2 and 3.3 and is tested by Travis. I don't use 3.2 nor 3.3 so we could remove them if it bothers us, until then… why not? WDY?T

harlowja:
Less versions of python that I don't have to think about the better :-P

jd:
Closed by #10

PR #10, opened 2016-08-17
Remove dead versions of py 3.x

harlowja:
The less pytho versions we have to
try to think about supporting (and
dealing with there edge-cases) the
better.

Fixes issue #4

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,6 @@ language: python
 python:
   - 2.6
   - 2.7
-  - 3.2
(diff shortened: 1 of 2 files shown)


jd:
You may want to disable it from Travis too then.

harlowja:
Pretty sure removing it from .travis.yml removes it from travis ;)

harlowja:
Ya, i'm also fine with removing 2.6 but this issue was about py3.x, haha

PR #13, opened 2016-08-18
Add  wait_chain()

bodenr:
One PR I had in retrying [1] proposed to support the notion of
"stepped wait". This patch adds stepping support by introducing
a new combined step function that wraps 1 or more wait fns to
produced a "stepped queue" of wait fns.

This approach solves [1] and also is reusable.

[1] https://github.com/rholder/retrying/pull/51

### DIFF
--- a/tenacity/wait.py
+++ b/tenacity/wait.py
@@ -69,6 +69,32 @@ class wait_combine(object):
             self.wait_funcs))
 
 
+class wait_chain(object):
+    """Chain two or more waiting strategies.
+
+    If all strategies are exhausted, the very last strategy is used
+    thereafter.
+
+    For example::
+
+        @retry(wait=wait_chain(*[wait_fixed(1000) for i in range(3)] +
+                               [wait_fixed(2000) for j in range(5)] +
+                               [wait_fixed(5000) for k in range(4)))
+        def wait_chained():
+            print("Wait 1s for 3 attempts, 2s for 5 attempts and 5s
+                   thereafter.")
+    """
+
+    def __init__(self, *strategies):
+        self.strategies = list(strategies)
+
+    def __call__(self, previous_attempt_number, delay_since_first_attempt_ms):
+        wait_func = self.strategies[0]
+        if len(self.strategies) > 1:
+            self.strategies.pop(0)
+        return wait_func(previous_attempt_number, delay_since_first_attempt_ms)
+
+
 class wait_incrementing(object):
     """Wait an incremental amount of time after each attempt.
 
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -120,6 +120,17 @@ services and other remote endpoints.
         print "Wait at least 3 seconds, and add up to 2 seconds of random delay"
 
 
+Sometimes it's necessary to build a chain of backoffs.
+
+.. code-block:: python
+
+    @retry(wait=wait_chain(*[wait_fixed(3000) for i in range(3)] +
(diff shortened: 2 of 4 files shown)


bodenr:
Not UTs yet; wanted to get your feedback on the approach here.

Let me know what you think and I'll proceed.

Thanks

tenacity/wait.py
```diff
@@ -69,6 +69,30 @@ def __call__(self, previous_attempt_number, delay_since_first_attempt_ms):
             self.wait_funcs))
 
 
+class wait_combine_stepped(object):
+    """Combine one or more waiting strategies in a stepped fashion.
+
+    For example::
+
+        @retry(wait=wait_combine_stepped(5, wait_fixed(1000), wait_fixed(2000), wait_fixed(5000)))
```
jd: Maybe i'd use the word "chain" rather than combine?

jd:
If we had a more general `wait_chain(*strategies)` class, you could do something along:

```
wait_chain([wait_fixed(5) for _ in range(5))] + [wait_fixed(10) for _ in range(5)])
```

For example. Not saying that a `wait_chain_stepped` is a no-go, but building it from `wait_chain` maybe funnier ;-)

bodenr:
@jd If you'd prefer me to update wait_combine() that's fine with me. I initially started hacking on it, but was killing the simplicity of the existing wait_combine() so I figured it would be cleaner to separate.

jd:
No I don't want to change `wait_combine`, I want a new method called `wait_chain` :)

tenacity/wait.py
```diff
@@ -69,6 +69,39 @@ def __call__(self, previous_attempt_number, delay_since_first_attempt_ms):
...
+    If all strategies are exhausted, the very last strategy is used
+    thereafter.
+
+    For example::
+
+        @retry(wait=wait_chain([wait_fixed(1000) for i in range(3)],
+                               [wait_fixed(2000) for j in range(5)],
+                               [wait_fixed(5000) for k in range(4)))
+        def wait_stepping():
+            print("Wait 1s for 3 attempts, 2s for 5 attempts and 4s thereafter.")
```
bodenr: Whoops. The text doesn't match the chain :) I can fix.

bodenr:
@jd something like this ^^ ?

tenacity/wait.py
```diff
@@ -69,27 +69,36 @@ def __call__(self, previous_attempt_number, delay_since_first_attempt_ms):
...
-    def __init__(self, step, *strategies):
-        self.step = int(step)
-        if self.step <= 0:
-            raise ValueError("Step '%s' must be larger than 0." % self.step)
-        self.wait_funcs = list(strategies)
+    def __init__(self, *strategies):
+        self.strategies = []
+        self._add_strategies(*strategies)
+
+    def _add_strategies(self, *strategies):
```
jd: Yeah this is a bit hackish, I think it's better to make the caller use + on its list so it can pass them as `wait_chain(*([mylist] + [mylist] …))` or just `wait_chain(wait_fixed(1000), wait_fixed(2000))`

bodenr:
@jd have a look now.

jd:
@bodenr This looks perfect. Can you fix the pep8 error and rebase/squash your branch into a proper clean commit? (or let me know I'll do it)

bodenr:
@jd let me know if ^ will suffice.

Thanks

jd:
LGTM!

PR #15, opened 2016-08-20
Add before/after strategies

harlowja:
Make these before and after functions more official
by providing some basic ones and a no-op ones that
will be used by default.

### DIFF
--- a/tenacity/__init__.py
+++ b/tenacity/__init__.py
@@ -43,6 +43,14 @@ from .wait import wait_jitter  # noqa
 from .wait import wait_none  # noqa
 from .wait import wait_random  # noqa
 
+# Import all built-in before strategies for easier usage.
+from .before import before_log  # noqa
+from .before import before_nothing  # noqa
+
+# Import all built-in after strategies for easier usage.
+from .after import after_log  # noqa
+from .after import after_nothing  # noqa
+
 from tenacity import _utils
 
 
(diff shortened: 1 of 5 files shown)


tenacity/after.py
```diff
@@ -0,0 +1,32 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from tenacity import _utils
+
+
+def nothing_after(func, trial_number):
```
jd: All functions so far are named `foo_bar()` from module `foo`, but here you used `bar_foo`. I know English is better read this way, but for the sake of consistency… ?

harlowja: Sure sure will change :-P

harlowja:
Ok should be good to go...

tenacity/__init__.py
```diff
@@ -122,15 +130,20 @@ def call(self, fn, *args, **kwargs):
...
+            trial_start_time = now()
+            if self.before is not None:
+                self.before(fn, attempt_number)
 
             fut = Future(attempt_number)
             try:
-                fut.set_result(fn(*args, **kwargs))
+                result = fn(*args, **kwargs)
+                trial_end_time = now()
+                fut.set_result(result)
```
jd: I think both L140 and L141 should be put in the `else` statement. I was already eager to do it last time, but the more I see thing in this `try` code that shouldn't raise `TryAgain` nor raise any `Exception` embedded, the more I am unhappy.

e.g. I don't want now() failing being a cause of "should I retry?" :-/

jd:
Minor thing that I missed but I'm ok to fix it after :)

tenacity/__init__.py
harlowja: Works for me.

harlowja:
Done.

PR #18, opened 2016-08-21
Add and use a nice ordinal func when logging before/after

harlowja:


jd:
lol can't you just use attempt number # ?

harlowja:
No, this thing is super awesome :-P

jd:
What I would not do to please you!

PR #20, opened 2016-08-29
Add built-in DEBUG logging

harlowja:


tenacity/_utils.py
```diff
@@ -36,6 +41,13 @@ def capture(fut, tb):
         fut.set_exception(tb[1])
 
 
+def pick_first_not_none(*args, **kwargs):
```
jd: pip install first

:-)

jd:
TRACE does not exist, I don't think it's a good idea to invent it. Just go ahead with DEBUG.

harlowja:
So I've been there, yes, TRACE does not exist, but if its DEBUG, the openstack folks will flip out at us (because lots of people run openstack at DEBUG level, mainly because there is no other useful way to figure out what's happening in the code otherwise).

tenacity/_utils.py
harlowja: Meh :-P

jd:
I think it's good to have this at DEBUG level anyway as if you run your service at that level to debug it, this is interesting option.

The rest of the "OpenStack" problem, well, they'll have to fix it in their projects. ;)

harlowja:
Okie dokie, I'll adjust it soon :-P

PR #27, opened 2016-09-15
Add support for reraise

bodenr:
Add support for reraise as per https://github.com/jd/tenacity/issues/24

PR #28, opened 2016-09-22
Fix reraise when no exception

bodenr:
https://github.com/jd/tenacity/pull/27 introduced a bug
whereupon if a retry callable doesn't raise and Retrying has
its self.reraise set to True, a TypeError will be raised because
we try to raise something that's not an exception class.

This fixes that issue and adds a UT for it.

[1] https://github.com/jd/tenacity/pull/27


jd:
@bodenr released 3.1.1 with that fix, thanks!

PR #29, opened 2016-09-26
Add retrying porting section in README

bodenr:
Updated readme with lessons learned from porting code that uses retrying in openstack to tenacity. This could be useful for others wanting to migrate.

### DIFF
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -196,6 +196,81 @@ we can reraise the last attempt's exception if needed:
     except MyException:
         # timed out retrying
 
+Porting from retrying to tenacity
+---------------------------------
+
+While tenacity is a fork of `Retrying`_, there are a few differences to be aware of:
+
+- While retrying uses milliseconds, tenacity uses seconds:
+
+.. code-block:: python
+
+
+    @retrying.retry(wait_fixed=2000)
+    def do_something_in_retrying():
+        print("Every 2 seconds with retrying")
+
+
+    @tenacity.retry(wait=tenacity.wait_fixed(2))
+    def do_something_in_tenacity():
+        print("Every 2 seconds with tenacity")
+
+- By default retrying raises the last exception raised by the retried callable (where applicable),
+  whereas tenacity raises a `RetryError`. To force tenacity to reraise the last exception
+  use the `reraise` kwarg.
+
+.. code-block:: python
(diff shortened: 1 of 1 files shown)


doc/source/index.rst
```diff
@@ -196,6 +196,81 @@ we can reraise the last attempt's exception if needed:
     except MyException:
         # timed out retrying
 
+Porting from retrying to tenacity
+---------------------------------
+
+While tenacity is a fork of `Retrying`_, there are a few differences to be aware of:
```
jd: can you replace "is a fork of" by something like "orginally forked from" ?

doc/source/index.rst
```diff
@@ -196,6 +196,81 @@ we can reraise the last attempt's exception if needed:
...
+    def do_something_in_tenacity():
+        raise MyError("Tenacity does not retry on exception")
+
+    @tenacity.retry(wait=tenacity.retry_any(
+        tenacity.retry_if_exception_type(Exception), # retry on exception
+        tenacity.retry_if_result(check_result)))
+    def do_something_in_tenacity_with_exception_retry():
+        raise MyError("Tenacity retries on any exception")
+
+- Tenacity uses `time.sleep` as a default in one of its kwargs. As a result consumers
```
jd: Nobody sane will need that. Only crazy people testing things they should not test would need that, so I don't think this paragraph is a good idea.

doc/source/index.rst
```diff
@@ -196,6 +196,81 @@ we can reraise the last attempt's exception if needed:
...
+  needing to patch tenacity's `time.sleep` (for example in unit test mocking)
+  must do so before tenacity is imported. For example in `__init__.py` of your top-level
+  unit test package:
+
+.. code-block:: python
+
+
```
bersace: spurious empty line ?

doc/source/index.rst
bodenr: I likely don't have the background in python you guys do, but people (mock) patching ```time.sleep``` is a reality regardless of if its "sane" or not. For that reason I'm not crazy about removing this section all together unless the code is updated to not use ```time.sleep``` in kwargs.

If someone else wants to drive this PR into something others agree to merge; please do.

jd:
No update for a while.

PR #31, opened 2016-10-24
Fix typo in retry_if_exception example

gdavoian:
Besides the change, which is trivial, I have an additional question: in the print statement within the only_raise_retry_error_when_not_io_error function (lines 143-145) it is mentioned, that the function wraps in RetryError any other errors, but it seems to be incorrect, isn't it? As my experiments show, errors are not wrapped. By the way, is there currently any way to use something similar to wrap_exception as in retrying.retry?

### DIFF
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -140,7 +140,7 @@ exceptions, as in the cases here.
     def might_io_error():
         print "Retry forever with no wait if an IOError occurs, raise any other errors"
 
-    @retry(retry_on_exception=retry_if_io_error)
+    @retry(retry=retry_if_exception(retry_if_io_error))
     def only_raise_retry_error_when_not_io_error():
         print "Retry forever with no wait if an IOError occurs, raise any other errors wrapped in RetryError"
 
@@ -171,7 +171,7 @@ We can also combine several conditions:
 Any combination of stop, wait, etc. is also supported to give you the freedom
 to mix and match.
 
-It's also possible to retry explicitely at any time by raising the `TryAgain`
+It's also possible to retry explicitly at any time by raising the `TryAgain`
 exception:
 
 .. code-block:: python


jd:
You seem right. We should probably implement some doctest at some point. :) Thanks @gdavoian!

You can use reraise=True to not wrap exceptions.

gdavoian:
Thanks for your reply. Yes, reraise=True works for not wrapping exceptions, but it seems that there is no way to always wrap exceptions (unlike in retrying), and reraise=False doesn't help here :)

jd:
@gdavoian exceptions are always wrapped by default. Do you have a test case you can share? Feel free to open a different issue. :)

gdavoian:
I am not sure whether it is worth opening a new issue yet :) 

Here is an example of code which shows differences between tenacity and retrying behaviors.

`import tenacity


@tenacity.retry(retry=tenacity.retry_if_exception_type(KeyError), reraise=True)
def a():
    1 / 0

# not wrapped
try:
    a()
except Exception as e:
    print(repr(e))


@tenacity.retry(retry=tenacity.retry_if_exception_type(KeyError), reraise=False)
def b():
    1 / 0

# not wrapped
try:
    b()
except Exception as e:
    print(repr(e))


import retrying


@retrying.retry(retry_on_exception=lambda e: isinstance(e, KeyError), wrap_exception=True)
def c():
    1 / 0

# wrapped
try:
    c()
except Exception as e:
    print(repr(e))


@retrying.retry(retry_on_exception=lambda e: isinstance(e, KeyError), wrap_exception=False)
def d():
    1 / 0

# not wrapped
try:
    d()
except Exception as e:
    print(repr(e))`

jd:
Ok I see what you mean now! So to me this is more a bug in retrying as it raises a `RetryError` even if the problem in the end has nothing to do with retrying anything, it just a generic error that does not match any of the retrying conditions.

So I don't see why you would want that in the end :)

PR #37, opened 2016-11-23
Retry coroutines

bersace:
Hi,

This PR ports https://github.com/rholder/retrying/pull/64 to tenacity. Thanks for the fork. The codebase is really clean.

It's a bit tricky to implement retry on coroutines. I did this by extracting the calling (aka attempt) and sleep logic from the retry control loop. This is done by using a generator. The control loop generate instructions (attempt, sleep or return), it's up to the wrapping code to do the actual IO, asynchronously or not.

I'm actually not very proud of this. Supporting old python versions and sync/async is a bit hackish. I'd be glad to hear your opinion on this contribution. Until it's mergeable, i wish :)

Regards,

### DIFF
--- a/tenacity/__init__.py
+++ b/tenacity/__init__.py
@@ -16,6 +16,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+try:
+    import asyncio
+except ImportError:
+    asyncio = None
+
 from concurrent import futures
 import sys
 import threading
@@ -71,9 +76,11 @@ def retry(*dargs, **dkw):
     if len(dargs) == 1 and callable(dargs[0]):
         return retry()(dargs[0])
     else:
-        r = Retrying(*dargs, **dkw)
-
         def wrap(f):
+            if asyncio and asyncio.iscoroutinefunction(f):
+                r = AsyncRetrying(*dargs, **dkw)
+            else:
+                r = Retrying(*dargs, **dkw)
 
             @six.wraps(f)
             def wrapped_f(*args, **kw):
@@ -281,3 +288,7 @@ class RetryError(Exception):
 
     def __str__(self):
         return "RetryError[{0}]".format(self.last_attempt)
+
+
+if asyncio:
+    from tenacity.async import AsyncRetrying
--- a/tenacity/async.py
+++ b/tenacity/async.py
@@ -1,0 +1,52 @@
+# -*- coding: utf-8 -*-
+# Copyright 2016 Étienne Bersac
+# Copyright 2016 Julien Danjou
+# Copyright 2016 Joshua Harlow
+# Copyright 2013-2014 Ray Holder
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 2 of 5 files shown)


jd:
@haypo I'd love to have your opinion on this patch!

bersace:
cc @jpic

bersace:
Grün :)

tenacity/__init__.py
```diff
@@ -55,6 +58,11 @@
 
 from tenacity import _utils
 
+if sys.version_info > (3, 4):
```
haypo: asyncio is part of Python stdlib since Python 3.4.

It is possible to install it as a third party module on Python 3.3, but I'm not sure that it's worth it to support Python 3.3 (it's no more used, and it wasn't never very popular).

bersace:
``` python
>>> (3, 4, 0) > (3, 4)
True
```

As you wish.

bersace: Pyt3.3 is not tested on travis for this project :)

tenacity/async.py
```diff
@@ -0,0 +1,43 @@
...
+# limitations under the License.
+
+import asyncio
+import sys
+
+import tenacity._utils as _utils
+
+
+@asyncio.coroutine
+def call(self, fn, *args, **kwargs):
```
bersace: Yep, `retrying` is good.

bersace:
@haypo do you mean the test case is not effective ?

bersace:
@harlowja @jd do you want an implementation without iterator ? That's a nice piece of refacto. I might do this in two distinct PR.

jd:
@bersace if you squash your fixup commits that will make that easier to review first.

If the code without iterator is simpler, give it a try in another PR, that might be interesting or help us review both PR :)

bersace:
@jd let's try this.

bersace:
@jd @haypo here you are :)

bersace:
@jd rebased. One-commit PR to add retry on coroutines :)

tenacity/async.py
```diff
@@ -0,0 +1,52 @@
...
+                    exc_info = None
+                    continue
+                except Exception:
+                    result = NO_RESULT
+                    exc_info = sys.exc_info()
+                    continue
+            elif isinstance(do, DoSleep):
+                result = NO_RESULT
+                exc_info = None
+                yield from asyncio.sleep(do)
```
jd: So you're really ignoring self.sleep?

bersace:
Green \o/

tenacity/async.py
bersace: `self.sleep` is not in `BaseRetrying` but (Sync)`Retrying`. If you want, i can add a `sleep=asyncio.sleep` `__init__()` argument.

jd: You're right, I forgot that. I don't think it makes sense to have a different sleep here since it's tied to asyncio, so that's good. :)

bersace:
@jd thanks. Do you plan a release soon ? :)

jd:
@bersace just pushed 3.6.0

PR #38, opened 2016-11-24
Align tox and travis settings

bersace:
I'm fighting with tox and travis in #37 . Let's clean this a bit here.


bersace:
@jd are you ok with that ? :)

PR #40, opened 2016-12-01
Split marshall of retries from IOs

bersace:


bersace:
@jd @haypo here you are.

tenacity/__init__.py
```diff
@@ -56,7 +54,12 @@
 from .after import after_log  # noqa
 from .after import after_nothing  # noqa
 
+from .marshaller import Future  # noqa
```
bersace: Actually, it looks like it's used only in tests. But just to *not touch tests* at all in this PR, I added this imports.

bersace:
@jd @haypo on dirait que ça ne vous motive pas beaucoup plus  ^^

jd:
@bersace I think I like the idea of an iterator actually. I just need to properly review the code now :)

bersace:
@jd cool :)

tenacity/__init__.py
jd: Good idea :) Let's keep it simple we can fix it later anyway

tenacity/marshaller.py
```diff
@@ -0,0 +1,149 @@
...
+            fut.set_result(value)
+        return fut
+
+
+class Marshaller(object):
+    def __init__(self, retrying, fn):
+        self.retrying = retrying
+        self.fn = fn
+
+        retrying.statistics.clear()
```
jd: I wonder at this stage if the statistics object should not be entirely managed by the Marshaller?

bersace:
@jd yes. It is a pure refacto (tests are not even touched). The purpose is just to decouple IO from marshalling of the retry loop. Making it possible to use the marshaller in a async wrapper.

jd:
So do you envision implementing a different Marshaller for asyncio? That's what would make sense to me. :)

tenacity/marshaller.py
bersace: Good point. But marshaller is a volatile objects while statistics must remain. I investigate this.

bersace:
Nope. Maybe the name is misleading. The marshaller is IO agnostic. It just returns actions (DoSleep, DoAttempt, etc.) or the actual value.

The retrying controller is `Retrying.call`. Do you want to add `Retrying.call_async` in this PR ?

The target API is to retry a?sync transparently.

``` python
@retry
def sync_method():
    do_io()

@retry
async def async_method():
    await do_aio()
````

**Comment by bersace** [comment]
I may also let `retry` decorator instanciate `RetryingAsync` on coroutine functions and `RetryingSync` otherwise. That should be cleaner.

**Comment by bersace** [comment]
I suggest de iterate on the sync/async controller implementation on the next PR.

**Comment by bersace** [comment]
@jd do you think `Policy` is better than `Marshaller` ?

**Comment by bersace** [comment]
@jd Here you are. The new version regroup all settings + marshall logic in Marshaller. The controller is responsible of looping and I/Os. What do you think of it ?

**Comment by bersace** [comment]
The PR becomes quite bigger :(

**Comment by jd** [comment]
It's hard to review if you don't squash commits :)

**Comment by bersace** [comment]
squashed

File: tenacity/__init__.py [hunk 3]
```diff
@@ -95,153 +83,32 @@ def wrapped_f(*args, **kw):
...
     """Retrying controller."""
 
-    def __init__(self,
-                 stop=stop_never, wait=wait_none(),
-                 sleep=time.sleep,
-                 retry=retry_if_exception_type(),
-                 before=before_nothing,
-                 after=after_nothing,
-                 reraise=False):
+    def __init__(self, marshaller, sleep=time.sleep):
```
jd: This is breaking the public API so this is not an option.

You should create the marshaller in `__init__` probably.

tenacity/__init__.py
```diff
@@ -95,153 +83,32 @@ def wrapped_f(*args, **kw):
...
-                  there own view of statistics they have collected (in the
-                  future we may provide a way to aggregate the various
-                  statistics from each thread).
-        """
-        try:
-            return self._local.statistics
-        except AttributeError:
-            self._local.statistics = {}
-            return self._local.statistics
+        self.statistics = marshaller.statistics
```
jd: Hm rather than copy the ref (I guess it works because it's a dict) maybe go for a `@property`.

tenacity/marshaller.py
```diff
@@ -0,0 +1,190 @@
...
+
+        return self._local.statistics
+
+    def __repr__(self):
+        attrs = _utils.visible_attrs(self, attrs={'me': id(self)})
+        return ("<Marshaller object at 0x%(me)x (stop=%(stop)s, "
+                "wait=%(wait)s, sleep=%(sleep)s, retry=%(retry)s, "
+                "before=%(before)s, after=%(after)s)>") % (attrs)
+
+    def begin(self, fn):
```
jd: Is there anything preventing to merge with `__init__`?

tenacity/__init__.py
bersace: Well, it does not break `retry` but `Retrying`. I feel `Retrying` as something private. Do you ?

tenacity/marshaller.py
bersace: I changed Marshaller to be instanciated per decorated function rather than per call. This is why i need to split the decorate-init (`__init__`) and the call-init (`begin`).g

bersace:
@jd actually, i find it quite nice to tests all behaviour with Marshaller rather than Retrying. You don't have to mock `time.sleep` anymore since IOs are separated.

bersace:
@jd Updated. Thanks for the review :)

tenacity/__init__.py
```diff
@@ -86,8 +86,8 @@ def wrapped_f(*args, **kw):
 class Retrying(object):
     """Retrying controller."""
 
-    def __init__(self, marshaller, sleep=time.sleep):
-        self.marshaller = marshaller
+    def __init__(self, sleep=time.sleep, **kwargs):
```
jd: That's good, but it kind of kills the documentation as we lose the argument list. Since it's the major exposure of the API, I think it'd be better to just duplicate the list with Marshaller.

jd:
A more general question: why in the end implement a `Marshaller` if there's only one anyway? it has all the logic that `Retrying` used to have split - which is good, I do understand the point. But is it necessary to go for another class?
(just a question, not a request to change)

tenacity/__init__.py
bersace: Ok, fix pushed.

bersace:
@jd the async PR will amend `retry` decorator to instanciate an async implementation of the controller, reusing the marshaller.

``` python
if asyncio.iscoroutinefunction(f):
    r = AsyncRetrying(*dargs, **dkw)
else:
    r = Retrying(*dargs, **dkw)
...
def wrapped_f(*args, **kw):
    return r.call(f, *args, **kw)

wrapped_f.retry = r
...
```

jd:
That's an interesting answer because it reinforces my view of having a base class `BaseRetrying` which can be inherited from both `Retrying` and `AsyncRetrying`. Which would make things way simpler and smoother that this marshaller thing. WDYT?

bersace:
@jd yes, the async will deduplicate `__init__` and `statistics` members in a `BaseRetrying`.

bersace:
Hmm, how do you split the files to support both python versions ?

```
__init__.py: from .async import AsyncRetrying
async.py: from tenacity import BaseRetrying
...
```

bersace:
Or something like:

``` python
from .sync import SyncController
from .async import AsyncController
```

```
sync.py: from .controller import BaseController
async.py: from .controller import BaseController
```

?

jd:
Yeah having 2 files sync and async and trying to import them and if it fails (like importing async on Py2) just fallback to always sync :) That sounds like a good idea.

bersace:
@jd i renamed it to `controller.py`. I think `retrying.py` would mislead to `retry.py`.

bersace:
Fixed flake8

bersace:
ok, green again.

jd:
That looks simpler indeed. I would not have moved things into controller though, I don't see the point of having yet another file for that. That's again breaking the API because Retrying is going away and **is** currently used by apps outside.

bersace:
@jd thanks for the review. `Retrying` compatiblity is restored. API is not broken by this PR. All is gathered in `__init__.py`

bersace:
@jd ok, i'm done. The PR should be minimal now.

tenacity/__init__.py
```diff
@@ -78,49 +81,47 @@ def wrapped_f(*args, **kw):
             wrapped_f.retry = r
             return wrapped_f
 
-        return wrap_simple(dargs[0])
+        return wrap
```
jd: I like the change but it has nothing to do with this PR unfortunately :(

tenacity/__init__.py
```diff
@@ -78,49 +81,47 @@ def wrapped_f(*args, **kw):
...
 
 
 class TryAgain(Exception):
     """Always retry the executed function when raised."""
 
 
-class Retrying(object):
-    """Retrying controller."""
+class BaseRetrying(object):
+    _local = threading.local()
```
jd: It's not clear to me that you need to change that.

bersace:
:up:

PR #43, opened 2016-12-13
Run flake8 only with latest python

bersace:
Hi,

I suggest to run flake8 with latest python version. This mean we do not use flake8 to check py2 compatibility, but only nosetests.

This allow to decouple validation of compatibility and code style.

Extracted from #37


jd:
Sounds good enough.

Issue #48, opened 2017-01-12
Provide a context manager


xmedeko:
I like the idea from https://github.com/bhearsum/redo to have `retrying` contextmanager:

```python
from tenacity import retry
from tenacity.stop import stop_after_attempt


@contextmanager
def retrying(func, *retry_args, **retry_kwargs):
    yield retry(*retry_args, **retry_kwargs)(func)


def foo(max_count):
    global count
    count += 1
    print(count)
    if count < max_count:
        raise ValueError("count too small")
    return "success!"

count = 0
ret = None
max_attempt=5
with retrying(foo, stop=stop_after_attempt(max_attempt)) as f:
    ret = f(3)
print(ret)
```

Do you think is may be included in the `tenacity`? I have no clue if it works with `async` and `futures`.

jd:
Yes, I think having a context manager would be an interesting feature. :)

jd:
So there's actually no interesting thing we can do with the `with` statement in Python. It'd be useful if there was some macro, but, nop.

Basically your example turns down to be just:

```python
f = tenacity.Retrying(stop=stop_after_attempt(max_attempt)
f.call(foo, 3)
```

I've added `__call__` so it can be reduced to:

```python
f = tenacity.Retrying(stop=stop_after_attempt(max_attempt)
f(foo, 3)
```

xmedeko:
+1 I am still novice in Python, so it seem's the `redo` solution is overkill. For me, the `f.call()` is well enough. IMHO it's just necessary to document such features.

proppy:
> So there's actually no interesting thing we can do with the with statement in Python. It'd be useful if there was some macro, but, nop.

What about something like this:
```
retrying = tenacity.Retrying(stop=stop_after_attempt(max_attempt))
with retrying:
    foo(3)
```

With something like this:
```
    def __exit__(self, exc_type, exc_val, exc_tb):
        if exc_type is None:
            self.attempt_number = 0
            return True
        else:
            self.attempt_number += 1
            return self.attempt_number <= self.max_attempt_number
```

jd:
@proppy sounds like a good idea. It's not much different than doing `retrying(foo, 3)` but I can see the value of being more Pythonic. Wanna do a PR? ;)

heri16:
I would rather the context manager behaves more pythonic:
```
with retrying(func, ...) as func_result:
  print(func_result.http_statuscode)
```

jd:
[...]

PR #59, opened 2017-04-22
Add RetryingContext class (A)

heri16:
This implementation does not support AsyncRetrying.
See Issue #48

Example:
```python
with RetryingContext(open) as open:
  with open('filename.txt', 'r') as file:
    ...
```

**Flow:** ContextManager(func, retry_opts) -> retriable_func(func_params) -> func_result -> ContextManager -> file

Outputs a context manager that returns `retriable_func` on `__enter__`.


jd:
This add 0 value to the workflow. It's just equivalent to

```python
f = retry(someargs)(f)
```

The with statement adds nothing interesting.

PR #60, opened 2017-04-22
Add RetryingContext class (B)

heri16:
Alternative implementation with support for AsyncRetrying.

See Issue #48

Example:
```python
with RetryingContext(open) as open:
  with open('filename.txt', 'r') as file:
    ...

with RetryingContext(asyncio.coroutine(open)) as open:
  with (yield from open('filename.txt', 'r')) as file:
     ...
```

**Flow:** ContextManager(func) -> retriable_func(func_args) -> func_result -> ContextManager -> file
**Async Flow:** ContextManager(async_func) -> async_retriable_func(func_args) -> func_result -> ContextManager -> file

Outputs a context manager that returns `retriable_func` or `async_retriable_func` on `__enter__`.
Awaiting of `func_result` from the `async_retriable_func` is managed externally by the user.

### DIFF
--- a/tenacity/__init__.py
+++ b/tenacity/__init__.py
@@ -264,6 +264,7 @@ class RetryingContext(object):
     else:
       r = Retrying(*self.args, **self.kwargs)
 
+    @six.wraps(f)
     def wrapped_fn(*args, **kw):
       return r.call(f, *args, **kw)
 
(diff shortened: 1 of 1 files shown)


jd:
The with statement adds nothing again here. It's just equivalent to `f = retry(someargs)(f)`.

PR #61, opened 2017-04-22
Add AsyncRetryingContext class (A)

heri16:
This implementation uses a single shared async context manager.
See Issue #48

**Example:**
```python
async with AsyncRetryingContext(open, wait=wait_fixed(2))('filename.txt', 'r') as file:
  ...
```

**Flow:** AsyncContextManager(func, retry_opts)(func_args) -> async_retriable_func(func_args) -> [func_result] -> file

Outputs an async context manager that generates `async_retriable_func` on `__aenter__`. It then internally passes in `func_args` into this `async_retriable_func` and awaits for the `func_result`.
- If the `func_result` is a value, this value is returned by by `__aenter__` to be passed into the new context as `file`.
- If the `func_result` is another context manager, it also calls `__enter__` on this wrapped context manager. If `__enter__` returns a truthy value, this value is in turn returned by `__aenter__` to be passed into the new context as `file`.

### DIFF
--- a/tenacity/async.py
+++ b/tenacity/async.py
@@ -50,3 +50,41 @@ class AsyncRetrying(BaseRetrying):
                 yield from asyncio.sleep(do)
             else:
                 return do
+
+class AsyncRetryingContext(AsyncRetrying):
+  """An asynchronous context manager is a context manager that is able to suspend execution in its enter and exit methods."""
+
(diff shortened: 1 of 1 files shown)


heri16:
Ok, this implementation has no merit. Closing.

PR #62, opened 2017-04-22
Add AsyncRetryingContext class (B)

heri16:
Alternative implementation that uses nested async context manager.
See Issue #48

**Example:**
```python
async def ticker(delay, to):
    """Yield numbers from 0 to `to` every `delay` seconds."""
    for i in range(to):
        yield i
        await asyncio.sleep(delay)

with AsyncRetrying(ticker, wait=wait_fixed(2)) as ticker:
    async for i in ticker(1, 10):
        print(i)
```

### DIFF
--- a/tenacity/async.py
+++ b/tenacity/async.py
@@ -19,6 +19,9 @@
 import asyncio
 import sys
 
+import six
+from contextlib
+
 from tenacity import BaseRetrying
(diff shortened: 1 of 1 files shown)


heri16:
No idea what was on my mind... we need an underlying support for retrying generators first. Closing...

PR #67, opened 2017-04-25
Wrap callable into coroutine in AsyncRetrying

heri16:
In the situation of having legacy libraries within an asyncio stack, it is more convenient for `AsyncRetrying` to check that a coroutine is passed in to avoid silly `object not iterable` errors.

### DIFF
--- a/tenacity/async.py
+++ b/tenacity/async.py
@@ -28,6 +28,8 @@ from tenacity import NO_RESULT
 class AsyncRetrying(BaseRetrying):
     @asyncio.coroutine
     def call(self, fn, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


jd:
I don't know enough of asyncio right now to validate. @bersace, any opinion?

bersace:
@heri16 Thanks for the contribution.

The thing is that `AsyncRetrying` is called **when** `fn` is an `asyncio.coroutine`. See https://github.com/jd/tenacity/blob/master/tenacity/__init__.py#L80 . Sorry, i don't understand :)  Can you point me a way to reproduce the `object not iterable` error ?

PR #70, opened 2017-06-08
feat: added wait_full_jitter

william-silversmith:
Hi,

I've implemented a wait strategy based on the results of this Amazon Architecture Blog:

https://www.awsarchitectureblog.com/2015/03/backoff.html

The Full Jitter strategy attempts to prevent synchronous clusters
from forming in distributed systems by combining exponential backoff
with random jitter. This differs from other strategies as jitter isn't
added to the exponentially increasing sleep time, rather the time is
computed is uniformly random between zero and the exponential point.

Excerpted from the above blog:

      sleep = random_between(0, min(cap, base * 2 ** attempt))

A competing algorithm called "Decorrelated Jitter" is competitive
and in some circumstances may be better. c.f. the above blog for
details.

This is a little bit of a hack because it would be possible to implement this by
making `wait_random` take a function or integer instead of only integers. That would
allow for composing `wait_exponential` and `wait_random` to achieve this effect.

Hope you find this helpful!
Will

### DIFF
--- a/tenacity/wait.py
+++ b/tenacity/wait.py
@@ -146,3 +146,40 @@ class wait_exponential(wait_base):
         except OverflowError:
             return self.max
         return max(0, min(result, self.max))
+
+
+class wait_full_jitter(wait_exponential):
+    """Random wait with exponentially widening window.
+
+    Wait strategy based on the results of this Amazon Architecture Blog:
+
+    https://www.awsarchitectureblog.com/2015/03/backoff.html
+
+    The Full Jitter strategy attempts to prevent synchronous clusters
+    from forming in distributed systems by combining exponential backoff
+    with random jitter. This differs from other strategies as jitter isn't
+    added to the exponentially increasing sleep time, rather the time is
+    computed is uniformly random between zero and the exponential point.
+
+    Excerpted from the above blog:
+        sleep = random_between(0, min(cap, base * 2 ** attempt))
+
+    A competing algorithm called "Decorrelated Jitter" is competitive
+    and in some circumstances may be better. c.f. the above blog for
+    details.
+
+    Example:
+        wait_full_jitter(0.5, 60) # initial window 0.5sec, max 60sec timeout
+
+    Optional:
+        base: (float) starting jitter window size in seconds.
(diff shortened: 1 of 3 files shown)


tenacity/wait.py
```diff
@@ -146,3 +146,49 @@ def __call__(self, previous_attempt_number, delay_since_first_attempt):
...
+
+        assert base >= 0
+        assert max_timeout_sec >= 0
+
+        self.base = base
+        self.cap = max_timeout_sec
+
+    def __call__(self, previous_attempt_number, delay_since_first_attempt):
+        high = min(self.cap, self.base * (2 ** previous_attempt_number))
+        return random.uniform(0, high)
```
jd: Can you implement this has a child class of `wait_exponential`? That'd make sure it uses the same code base, voiding some bugs like catching `OverflowError`.

william-silversmith:
Hi Julian, 

What do you think of this iteration? I tried composing the functions instead of subclassing. Funnily enough, I was able to get the tests mostly passing with this simple, but more confusing, code. 

```
def wait_full_jitter(base, cap): 
    exp = wait_exponential(base, cap)
    return lambda pan, dsfa: wait_random(0, exp(pan, dsfa))(pan, dsfa)
```

jd:
If you use subclassing the whole code here:

```python
 def __init__(self, base=0.5, max_timeout_sec=60.0):
         super(self.__class__, self).__init__()
 
         assert base >= 0
         assert max_timeout_sec >= 0
 
         self.exp = wait_exponential(multiplier=base, max=max_timeout_sec)
 
     def __call__(self, previous_attempt_number, delay_since_first_attempt):
         high = self.exp(previous_attempt_number, delay_since_first_attempt)
         return random.uniform(0, high)
```
becomes

```python
 def __call__(self, previous_attempt_number, delay_since_first_attempt):
         return random.uniform(0, super(wait_full_jitter, self).__call__(previous_attempt_number, delay_since_first_attempt))
```

Hard to make that simpler. And you get the overflow error handling for free.

william-silversmith:
Ok! I made the requested changes. :D

PR #71, opened 2017-06-13
Experiment: Functionalizing all Arguments

william-silversmith:
Here's an experiment that might be interesting to you. This PR isn't cleaned up for merge, but it shows a little of what you can do if all wait functions take a function as an argument.

I've implemented the `wait_full_jitter` function from #70 as composed functions. It's a little messy, but mostly because of the length of the names of `previous_attempt_number` and `delay_since_first_attempt`. 

Let me know what you think.
- Will

### DIFF
--- a/tenacity/__init__.py
+++ b/tenacity/__init__.py
@@ -51,6 +51,7 @@ from .wait import wait_chain  # noqa
 from .wait import wait_combine  # noqa
 from .wait import wait_exponential  # noqa
 from .wait import wait_fixed  # noqa
+from .wait import wait_full_jitter  # noqa
 from .wait import wait_incrementing  # noqa
 from .wait import wait_none  # noqa
 from .wait import wait_random  # noqa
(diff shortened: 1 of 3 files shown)


jd:
Seems like complicating the code for no gain. If you don't want you min to be 1 but the result of a function… just call that function outside tenacity and passes the result? :)

william-silversmith:
Thanks for taking a look! This might be overcomplicating things, but I'm not sure I follow your critique. Would you mind showing a short example? In this case, it's not that we want min to be the result of a single call of the function, but we want min to change with every advance of the clock. 

The other way to do it is to do what we were discussing in #70, which isn't a huge deal, but it does take a little understanding of tenacity internals to set it up right and support, e.g. wait_combine. If you pass a function as an argument it'll 'just work'. Of course, it's also counterintuitive that the arguments can be functions, so one might argue that a similar level of understanding is required.

william-silversmith:
On the other hand, I do like your version in the comments on #70.

PR #79, opened 2017-07-07
Retry until exception type

Brian-Williams:
Add complement for retry_if_exception_type, which performs the inverse operation of ignoring all results and exceptions until an exception is hit.

Feel free to change name of the class. Also, I understand if this is superfluous and is closed. I needed it in a very specific use case and it may not be useful to others.

### DIFF
--- a/tenacity/retry.py
+++ b/tenacity/retry.py
@@ -64,12 +64,27 @@ class retry_if_exception(retry_base):
 
 
 class retry_if_exception_type(retry_if_exception):
-    """Retries if an exception has been raised of a certain type."""
+    """Retries if an exception has been raised of one or more types."""
 
     def __init__(self, exception_types=Exception):
         self.exception_types = exception_types
         super(retry_if_exception_type, self).__init__(
             lambda e: isinstance(e, exception_types))
+
+
+class retry_unless_exception_type(retry_if_exception):
+    """Retries until an exception is raised of one or more types."""
+
+    def __init__(self, exception_types=Exception):
+        self.exception_types = exception_types
+        super(retry_unless_exception_type, self).__init__(
+            lambda e: not isinstance(e, exception_types))
+
+    def __call__(self, attempt):
+        # always retry if no exception was raised
+        if not attempt.failed:
+            return True
+        return self.predicate(attempt.exception())
 
 
 class retry_if_result(retry_base):
--- a/tenacity/tests/test_tenacity.py
+++ b/tenacity/tests/test_tenacity.py
@@ -382,7 +382,7 @@ class NoneReturnUntilAfterCount(object):
 
 
 class NoIOErrorAfterCount(object):
-    "Holds counter state for invoking a method several times in a row."
+    """Holds counter state for invoking a method several times in a row."""
 
     def __init__(self, count):
         self.counter = 0
@@ -400,7 +400,7 @@ class NoIOErrorAfterCount(object):
 
 
 class NoNameErrorAfterCount(object):
-    "Holds counter state for invoking a method several times in a row."
+    """Holds counter state for invoking a method several times in a row."""
 
     def __init__(self, count):
         self.counter = 0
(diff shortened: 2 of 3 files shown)


tenacity/retry.py
```diff
@@ -64,14 +64,26 @@ def __call__(self, attempt):
...
 
+class retry_until_exception_type(retry_if_exception):
+    """Retries until an exception is raised of one or more types."""
+
+    def __init__(self, exception_types=Exception):
+        self.exception_types = exception_types
+        super(retry_until_exception_type, self).__init__(
+            lambda e: not isinstance(e, exception_types))
+
+    def __call__(self, attempt):
```
jd: Why do you need to override that?

tenacity/tests/test_tenacity.py
```diff
@@ -382,7 +382,7 @@ def go(self):
 
 
 class NoIOErrorAfterCount(object):
-    "Holds counter state for invoking a method several times in a row."
+    """Holds counter state for invoking a method several times in a row."""
```
jd: Would be great to have that in a different commit/PR.

tenacity/tests/test_tenacity.py
```diff
@@ -409,14 +409,50 @@ def __init__(self, count):
     def go(self):
         """Raise a NameError until after count threshold has been crossed.
 
-        Tthen return True.
+        Then return True.
```
jd: ditto.

tenacity/retry.py
Brian-Williams: I want to retry even if no exception is raised.

I'm not sure why `retry_if_exception` needs the conditional. I would've simply removed it there, but didn't want to alter code I didn't dive deeply into. Is it an optimization?

Brian-Williams: If it is an optimization a similar version for this class would be:

```
def __call__(self, attempt):
    # always retry if no exception was raised
    if not attempt.failed:
        return True
    return self.predicate(attempt.exception())
```

tenacity/tests/test_tenacity.py
Brian-Williams: I normally would separate the unique changes like I did in #80, but I don't like making style only pull requests as it simply makes git blame harder to read.

Those style changes make sense in that commit IMO. Since the docstrings were copied in the complement functions, we made their style match. And I didn't alter anything I didn't directly copy/change. For example there is still non-triple quoted docstrings for `NoneReturnUntilAfterCount`.

If you prefer style only pull requests I can split it off. If I wasn't adding `IOErrorUntilCount` with a matching docstring to this function I would've never changed this.

tenacity/tests/test_tenacity.py
Brian-Williams: This shouldn't be in this pull request. Apologies.

Brian-Williams:
Thanks for the quick response, jd! :D

I rebased to 2 commits, one adding the functionality and one testing it.

tenacity/retry.py
```diff
@@ -64,14 +64,26 @@ def __call__(self, attempt):
...
-    """Retries if an exception has been raised of a certain type."""
+    """Retries if an exception has been raised of one or more types."""
 
     def __init__(self, exception_types=Exception):
         self.exception_types = exception_types
         super(retry_if_exception_type, self).__init__(
             lambda e: isinstance(e, exception_types))
 
 
+class retry_until_exception_type(retry_if_exception):
```
jd: It's generally admitted that
* `if not` == `unless`
* `while not` == `until`

Since  it's basically the opposite of `retry_if_exception_type` I'd suggest renaming to `retry_unless_exception_type`.

tenacity/retry.py
jd: I'd like this optimization. I think it's a good call to not run the predication function, that could be heavy, if no need to.

tenacity/tests/test_tenacity.py
jd: > but I don't like making style only pull requests as it simply makes git blame harder to read.

What makes git blame hard to read is the change itself, not that it is a specific PR.

But it's not big deal. I wrote "would be great" so I can pinch my nose ;)

tenacity/tests/test_tenacity.py
jd: Don't worry, it's no big deal.

tenacity/retry.py
Brian-Williams: Makes sense to me. Hardest thing in programming for me is naming stuff.

Brian-Williams:
Renamed, optimized and squashed into one commit with bullet points. AFAIK this is good to merge. Feel free to ping for any other needed changes.

PR #82, opened 2017-07-25
Document random exponential backoff strategy

zaneb:
The `wait_full_jitter` strategy is very widely useful (it turns out that I was [implementing the same thing](https://review.openstack.org/#/c/328613/12/heat/engine/sync_point.py) at roughly the same time as the author), but it is not documented in the README and the name requires context to make sense of.

This PR renames the strategy to the hopefully self-explanatory `wait_random_exponential` and documents it better.


jd:
@zaneb thanks! makes sense. Released with 4.4.0. Enjoy!

PR #89, opened 2017-09-21
Issue-8: Make sure doc examples run

vicyap:
I am addressing issue #8. I added the sphinx doctest extension.

There was one case I left out, namely the async case, because I am not sure
how to get doctest to run on specific python versions without using
`.. doctest` blocks.


doc/Makefile
```diff
@@ -0,0 +1,130 @@
+# Makefile for Sphinx documentation
```
jd: We don't need this file, just remove it.

requirements_dev.txt
```diff
@@ -0,0 +1,2 @@
+Sphinx==1.6.3
```
jd: ditto, you can remove this file.

tox.ini
```diff
@@ -11,10 +11,13 @@ pypy = pypy
...
 usedevelop = True
 sitepackages = False
-deps = nose
+deps =
+    nose
+    sphinx
 commands =
     py{27,py}: python setup.py nosetests --ignore-files '.*async.py'
     py3{4,5,6}: python setup.py nosetests
+    py{27,34,35,36,py}: /usr/bin/make -C doc doctest
```
jd: just use `python setup.py build_sphinx`

and add to `setup.cfg`

```
[build_sphinx]
builders = html,doctest
```

Issue #101, opened 2017-12-01
Ability to have waiters know about prior runs result


harlowja:
I am using tenacity in a slack bot that is being used; and I'm using it around the code that sends all the things to slack apis; those apis sometimes report back issues that should be retried (as would be expected).

One of the interesting ones is when they send back the response 'rate_limited' :

*   https://api.slack.com/docs/rate-limits

When they do this they will send back a `Retry-After` header (which there python client now includes in there custom exception class) so it would be nice to have a waiter class that can be given the prior result (or exception) and adjust the amount of time waited based on that result.

Right now it seems that waiters just get a `previous_attempt_number, delay_since_first_attempt` which isn't quite enough to extract the suggested  `Retry-After` header from. 😢 

So might need a new waiter base class that actually gets the prior result...

PR #102, opened 2017-12-01
Allow waiters to introspect last result

harlowja:
This will help implement the desire listed at issue #101
by allowing for waiter functions/classes/other to recieve
the last result and do smart things with it.

### DIFF
--- a/tenacity/__init__.py
+++ b/tenacity/__init__.py
@@ -27,6 +27,7 @@ try:
 except ImportError:
     tornado = None
 
+import inspect
 import sys
 import threading
 from concurrent import futures
@@ -36,6 +37,7 @@ from monotonic import monotonic as now
 import six
 
 from tenacity import _utils
+from tenacity import wait as _wait
 
 # Import all built-in retry strategies for easier usage.
 from .retry import retry_all  # noqa
@@ -138,6 +140,10 @@ class BaseRetrying(object):
         self.after = after
         self.reraise = reraise
         self._local = threading.local()
+        # This will allow for passing in the result and handling
+        # the older versions of these functions that do not take
+        # the prior result.
+        self._wait_takes_result = self._waiter_takes_last_result(wait)
 
     def copy(self, sleep=_unset, stop=_unset, wait=_unset,
              retry=_unset, before=_unset, after=_unset, reraise=_unset):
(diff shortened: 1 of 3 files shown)


harlowja:
This doesn't need to be the way we do this, but this is one of such ways...

tenacity/__init__.py
```diff
@@ -251,8 +251,15 @@ def iter(self, result, exc_info, start_time):
...
         if self.wait:
-            sleep = self.wait(self.statistics['attempt_number'],
-                              delay_since_first_attempt)
+            # This will allow for passing in the result and handling
+            # the older versions of these functions that do not take
+            # the prior result.
+            try:
+                sleep = self.wait(self.statistics['attempt_number'],
+                                  delay_since_first_attempt, last_result=fut)
+            except TypeError:
```
harlowja: I don't like this capture here.... other option would be using `isinstance` of a new baseclass...

jd: did you try to `inspect` the signature?

harlowja: I didn't though that should work.

tenacity/__init__.py
```diff
@@ -80,6 +82,15 @@
 from .after import after_nothing  # noqa
 
 
+def _waiter_takes_last_result(waiter):
```
jd: this could/should be a method

tenacity/__init__.py
```diff
@@ -80,6 +82,15 @@
...
+    if not six.callable(waiter):
+        return False
+    if isinstance(waiter, _wait.wait_base):
+        waiter = waiter.__call__
```
jd: Not sure that's gonna work for code that inherits from `wait_base` but is not updated.

harlowja: Seems like it should work?

```
>>> from tenacity import wait
>>> class OldWait(wait.wait_base):
...    def __call__(self, attempt, delay):
...        pass
...
>>> my_wait = OldWait()
>>>
>>> import inspect
>>> inspect.getargspec(my_wait.__call__)
ArgSpec(args=['self', 'attempt', 'delay'], varargs=None, keywords=None, defaults=None)
```

tenacity/__init__.py
harlowja: Done

tenacity/__init__.py
jd: Sorry I misread the line 89. I read `waiter = wait_base.__call__`. Facepalm.

PR #109, opened 2018-03-15
Attach loop to the sleep

junqed:
Some testing libraries create their own loops and thus some tests could fail

### DIFF
--- a/tenacity/async.py
+++ b/tenacity/async.py
@@ -31,9 +31,11 @@ class AsyncRetrying(BaseRetrying):
 
     def __init__(self,
                  sleep=asyncio.sleep,
+                 loop=None,
                  **kwargs):
         super(AsyncRetrying, self).__init__(**kwargs)
         self.sleep = sleep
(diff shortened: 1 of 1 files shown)


sileht:
You can acheive the same thing by using `sleep=functools.partial(asyncio.sleep, loop=myloop)`

tenacity/async.py
```diff
@@ -58,6 +60,6 @@ def call(self, fn, *args, **kwargs):
             elif isinstance(do, DoSleep):
                 result = NO_RESULT
                 exc_info = None
-                yield from self.sleep(do)
+                yield from self.sleep(do, loop=self.loop)
```
jd: The problem is that if anyone was using a different sleep function that does not accept a loop argument, that will break them. As @sileht said, you can pass a lambda or a `functools.partial`

jd:
Closing as I don't think we can merge without taking a risk and it's not needed. Let me know if I missed something.

Issue #77, opened 2017-06-22
Improve on-retry logging


immerrr:
Hi!

And thank you for forking the one and only library for better retries in Python. Tenacity API looks better, although I'm still to find an opportunity to use it in my projects.

More often than not, if there's a retry, I'd like it logged somewhere. With `retrying` I found myself using `retry_on_exception` functions that contained log messages with smth like
```
X failed, retrying (maybe): <error message>
```
because `retry_on_exception` function doesn't know whether or not there will be another retry attempt.

tenacity makes a step in the right direction with `after=after_log` function, but it's kind of lacking because different bits of information that can be useful to log are available in different places, such as:
- first attempt start time or total time elapsed since then: available to `stop` functions
- next attempt delay: available as a return value of  `wait` functions
- attempt number: available to all (?) but `retry` functions
- invalid value returned or exception raised: available to `retry` functions

Providing just one logger/level/msgformat combination might not be enough as the user could want to use different levels for different situations, including:
- reporting retried attempts as WARNING and the last one as ERROR
- using different loglevels for different errors
- provide more information about the error in the log where possible

So I presume it should be another kwarg that accepts a function, e.g. `log_retry=` or `on_retry=`. And that function could accept an `attempt` with all interesting metadata attached (right now only `attempt_number` is attached to the `Future` instance), so as to avoid function signature issues when available metadata is altered.

jd:
I guess you can more or less do everything here already, but I admit it might not be obvious.

The last thing I want is a gigantic patch for this use case, but if you can enhance the current code and function and just add generic stuff so that's possible to do, give it a try. :)

lordi:
I'd like to have more advanced logging or more generalized before/after callbacks.

It would be nice to:

 * Access exception information (``exc_info``) in the ``after`` callback, if any.
 * Access all function arguments in ``after``. Right now, only the function itsself is given.

djmarcin:
The current state is sort of awkward for several reasons:
- The retry may not be configured to sleep so the name is unintuitive
- Assumes that an exception will be the cause of retries, but you can change this (e.g. `retry=retry_if_result(is_false_p)`). It doesn't crash, but it does print the nonsensical message `Retrying __main__.do_something_unreliable in 0 seconds as it raised None.` which misidentifies the reason for the retry.
- The final exception before a stop isn't printed, but it is reraised so not really a deal-breaker, just awkward because the logging won't be consistent.
- The traceback isn't printed which is often really useful.

[...]

PR #118, opened 2018-04-18
Implement before_sleep logging hook

lordi:
This PR arose from the wish to have the result and the future wait seconds in the ``after`` logging handler. It turned out that modifying this handler is not straight-forward because the wait seconds will only be calculated later.

To ensure backwards compatibility wrt to the ``wait`` function, I propose another handler ``before_sleep`` that will be called whenever a sleep is scheduled and the function will be retried later. For me, this logging hook is essential because I can be informed about the seconds tenacity will wait before retrying.

Related to jd/tenacity#77

### DIFF
--- a/tenacity/__init__.py
+++ b/tenacity/__init__.py
@@ -78,6 +78,10 @@ from .before import before_nothing  # noqa
 from .after import after_log  # noqa
 from .after import after_nothing  # noqa
 
+# Import all built-in after strategies for easier usage.
+from .before_sleep import before_sleep_log  # noqa
+from .before_sleep import before_sleep_nothing  # noqa
+
 
 def retry(*dargs, **dkw):
     """Wrap a function with a new `Retrying` object.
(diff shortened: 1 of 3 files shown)


tenacity/__init__.py
```diff
@@ -295,6 +305,10 @@ def iter(self, result, exc_info, start_time):  # noqa
         self.statistics['idle_for'] += sleep
         self.statistics['attempt_number'] += 1
 
+        if self.before_sleep is not None:
+            self.before_sleep(self.fn, self.statistics['attempt_number'],
```
jd: I wonder if it would not be simple to simple call `self.before_sleep(self, sleep=sleep, last_result=fut)` so you just have access to everything?

lordi: Yea, that make the hook simpler and more general. I'll adapt the PR.

lordi:
Done :-)

PR #119, opened 2018-04-29
Add optional retry_error_callback param

dbwpe:
Being able to call a custom function when all retries have failed is useful for a few applications.

* Do nothing (raise no exception without repeated try/except boilerplate)
* Return a default value
* Return the value of the last attempt (if `retry_if_result` for example)
* Really anything you can think to do with the last `Future` object

A couple of these uses were issues on the `retrying` module, which I made a PR to before realizing this is probably a better place for it.

### DIFF
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -254,6 +254,25 @@ In the same spirit, It's possible to execute after a call that failed:
     def raise_my_exception():
         raise MyException("Fail")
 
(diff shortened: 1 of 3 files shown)


dbwpe:
Hey there @jd - looks like you addressed the mergify failure in 1c0b5723126 :)

Any feedback on this PR that I can address?  (aside from merging in latest master)

PR #121, opened 2018-05-09
Improve docs

foobacca:
A couple of small things to make the docs easier to read.


foobacca:
Apparently I misunderstood how to add python syntax to the doc examples, so I've reverted the second commit.

PR #124, opened 2018-06-27
Make before_sleep accept call_state

immerrr:
This PR continues work started in #122 adding support for call_state in `before_sleep` and opening a possibility to improve logging even further as requested in comments in #77.

Before this is getting merged I'd like to update the documentation and fix `before_sleep_log` function to log non-exceptions as well.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,5 @@ AUTHORS
 ChangeLog
 .eggs/
 doc/_build
+
+/.pytest_cache
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -272,6 +272,109 @@ In the same spirit, It's possible to execute after a call that failed:
     def raise_my_exception():
         raise MyException("Fail")
 
+It's also possible to only log failures that are going to be retried. Normally
+retries happen after a wait interval, so the keyword argument is called
+``before_sleep``:
+
+.. testcode::
+
+    logger = logging.getLogger(__name__)
+    @retry(stop=stop_after_attempt(3),
+           before_sleep=before_sleep_log(logger, logging.DEBUG))
+    def raise_my_exception():
+        raise MyException("Fail")
+
+You can also define a custom ``before_sleep`` function. It should have one
+parameter called ``call_state`` that contains all information about current
+retry invocation.
+
+.. testcode::
+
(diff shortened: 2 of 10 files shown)


immerrr:
This is no longer "WIP", I've added some documentation and tweaked before_sleep_log to also print information about invalid results.

Given how natural it is to group `before`, `after` and `before_sleep` I think it could be better to tweak all three to accept `call_state` together.

Also, I wonder if `call_state` should be called `retry_state` until it's not part of the public interface. When I was patching this initially the `call_state` name made perfect sense, but after having left this alone for a while I'm not entirely convinced. WDYT?

.gitignore
```diff
@@ -8,3 +8,5 @@ AUTHORS
 ChangeLog
 .eggs/
 doc/_build
+
```
jd: empty line

doc/source/index.rst
```diff
@@ -272,6 +272,109 @@ In the same spirit, It's possible to execute after a call that failed:
...
+parameter called ``call_state`` that contains all information about current
+retry invocation.
+
+.. testcode::
+
+    logger = logging.getLogger(__name__)
+
+    def my_before_sleep(call_state):
+        if call_state.attempt_number < 1:
+            loglevel = logging.INFO
```
jd: logfn = logger.info

doc/source/index.rst
```diff
@@ -272,6 +272,109 @@ In the same spirit, It's possible to execute after a call that failed:
...
+        else:
+            loglevel = logging.WARNING
+        logger.log(
```
jd: `logfn(…)`

immerrr:
Hmm, I think mergify was a bit too swift to merge and close the PR for me to act on the comments :)

immerrr:
@jd any thoughts on the `call_state`/`retry_state` question?

jd:
I wrote the comments while reviewing the patches, but I did not think it was worth any update at this stage so I went ahead and approved, that's why it merged. :)

I forgot to reply though. So yeah I think all function should take a `call_state` as argument, so we're sure everybody gets as much information as possible so we're not limiting extendability.

`retry_state` sounds like a better named to export it, since yeah, `call_state` is a bit too much generic.

Issue #125, opened 2018-07-09
Add retry_if_exception_match


PhML:
As there is `retry_if_exception_type`, would it be possible to have `retry_if_exception_match`?
It would replace :
```python
retry_if_exception(lambda e: re.match(pattern, str(e)))
```
by:
```python
retry_if_exception_match(pattern)
```

immerrr:
It could be nicer to extend the existing `retry_if_exception` to accept `match` kwarg, like [pytest.raises does](https://docs.pytest.org/en/latest/assert.html#assertions-about-expected-exceptions):
```
retry_if_exception(MyException, match='zzz')
retry_if_exception(MyException, match=some_precompiled_regexp)
```

Brian-Williams:
I think having a separate classes is a good idea because it gives greater granularity for use cases where several retry predicates need to be chained.

It allows you to do things like separating exception type from exception message.

`retry=(retry_if_exception_type=MyException1 | retry_if_exception_message(match=r'a[bc]*'))`

Then even if exception `MyException2` raises with errormessage 'ab' it will retry.

@jd Any interest in this? I put together a rough sketch of what it could look like [here](https://github.com/Brian-Williams/tenacity/commit/8d19588434e0ea90bb83f56615153adba617c9ea).

jd:
@Brian-Williams That sounds like a good idea. Feel free to send a PR.

jamesbraza:
Should this issue be closed? I think it was resolved

PR #127, opened 2018-07-16
Add retry_if_exception_message

Brian-Williams:
From issue #125.

Add retry classes for exception messages.

I'm very open to name suggestions. Specifically wondering if `retry_if_not_exception_message` or `retry_until_exception_message` is best.

`retry_if_exception_message` and it's complement act different out of necessity. The normal class can end if there is no exception whereas it's inverse needs to retry if there is no exception message.

It's worth noting in my branch unit tests are intermittently failing on [wait class backwards incompatibility](https://travis-ci.org/Brian-Williams/tenacity/jobs/405236070#L470). I don't believe this is caused by any of my changes, but it's worth noting.

### DIFF
--- a/tenacity/retry.py
+++ b/tenacity/retry.py
@@ -15,6 +15,7 @@
 # limitations under the License.
 
 import abc
+import re
 
 import six
 
@@ -109,6 +110,50 @@ class retry_if_not_result(retry_base):
     def __call__(self, attempt):
         if not attempt.failed:
             return not self.predicate(attempt.result())
+
+
+class retry_if_exception_message(retry_if_exception):
+    """Retries if an exception message equals or matches."""
+
+    def __init__(self, message=None, match=None):
+        if message and match:
+            raise TypeError(
+                "{}() takes either 'message' or 'match', not both".format(
+                    self.__class__.__name__))
+
+        # set predicate
+        if message:
+            def message_fnc(exception):
+                return message == str(exception)
+            predicate = message_fnc
+        elif match:
+            prog = re.compile(match)
+
+            def match_fnc(exception):
+                return prog.match(str(exception))
+            predicate = match_fnc
+        else:
+            raise TypeError(
+                "{}() missing 1 required argument 'message' or 'match'".
+                format(self.__class__.__name__))
+
+        super(retry_if_exception_message, self).__init__(predicate)
+
+
+class retry_if_not_exception_message(retry_if_exception_message):
+    """Retries until an exception message equals or matches."""
+
+    def __init__(self, *args, **kwargs):
+        super(retry_if_not_exception_message, self).__init__(*args, **kwargs)
+        # invert predicate
+        if_predicate = self.predicate
+        self.predicate = lambda *args_, **kwargs_: not if_predicate(
+            *args_, **kwargs_)
+
+    def __call__(self, attempt):
+        if not attempt.failed:
+            return True
+        return self.predicate(attempt.exception())
 
 
 class retry_any(retry_base):
--- a/tenacity/__init__.py
+++ b/tenacity/__init__.py
@@ -47,6 +47,8 @@ from .retry import retry_if_not_result  # noqa
 from .retry import retry_if_result  # noqa
 from .retry import retry_never  # noqa
 from .retry import retry_unless_exception_type  # noqa
+from .retry import retry_if_exception_message  # noqa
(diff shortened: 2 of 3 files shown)


tenacity/retry.py
```diff
@@ -111,6 +112,61 @@ def __call__(self, attempt):
...
+            prog = re.compile(match)
+
+            def match_fnc(exception_message):
+                return prog.match(str(exception_message))
+            return match_fnc
+        raise TypeError(
+            "{}() missing 1 required argument 'message' or 'match'".format(
+                self.__class__.__name__))
+
+    def _no_failure(self):
```
jd: `@staticmethod`

Not sure it's worth having a method, just an attribute would be enough.

tenacity/retry.py
```diff
@@ -111,6 +112,61 @@ def __call__(self, attempt):
...
+        predicate = super(
+            retry_if_not_exception_message, self
+        )._predicate_factory(*args, **kwargs)
+
+        def inverted_predicate(*args_, **kwargs_):
+            return not predicate(*args_, **kwargs_)
+
+        return inverted_predicate
+
+    def _no_failure(self):
```
jd: @staticmethod

Not sure it's worth having a method, just an attribute would be enough.

tenacity/retry.py
```diff
@@ -111,6 +112,61 @@ def __call__(self, attempt):
...
+class retry_if_exception_message(retry_base):
+    """Retries if an exception message equals or matches."""
+
+    def __init__(self, message=None, match=None, require_exception=False):
+        if message and match:
+            raise TypeError(
+                "{}() takes either 'message' or 'match', not both".format(
+                    self.__class__.__name__))
+        self.predicate = self._predicate_factory(message, match)
+        self.require_exception = require_exception
```
jd: That does not seem to be related to the heart of the function. I'd remove that, and put that in a different class (and PR) if that's needed.

tenacity/retry.py
```diff
@@ -111,6 +112,61 @@ def __call__(self, attempt):
             return not self.predicate(attempt.result())
 
 
+class retry_if_exception_message(retry_base):
```
jd: you should inherit from `retry_if_exception`

tenacity/retry.py
```diff
@@ -111,6 +112,61 @@ def __call__(self, attempt):
...
+            return match_fnc
+        raise TypeError(
+            "{}() missing 1 required argument 'message' or 'match'".format(
+                self.__class__.__name__))
+
+    def _no_failure(self):
+        """Retry logic for no failure."""
+        return False
+
+    def __call__(self, attempt):
```
jd: This is not needed if you inherit from `retry_if_exception`

tenacity/retry.py
```diff
@@ -111,6 +112,61 @@ def __call__(self, attempt):
...
 
+class retry_if_exception_message(retry_base):
+    """Retries if an exception message equals or matches."""
+
+    def __init__(self, message=None, match=None, require_exception=False):
+        if message and match:
+            raise TypeError(
+                "{}() takes either 'message' or 'match', not both".format(
+                    self.__class__.__name__))
+        self.predicate = self._predicate_factory(message, match)
```
jd: just pass the result to the `__init__` method of `retry_if_exception` here (since you will inherit from it). The factory method is not even really needed IMHO, you can put that code in `__init__`.

tenacity/retry.py
```diff
@@ -111,6 +112,61 @@ def __call__(self, attempt):
...
+        if not attempt.failed:
+            if self.require_exception:
+                raise RuntimeError("No exception to get errormessage from")
+            return self._no_failure()
+        return self.predicate(attempt.exception().value)
+
+
+class retry_if_not_exception_message(retry_if_exception_message):
+    """Retries until an exception message equals or matches."""
+
```
jd: just set invert the predicate in `__init__` with a lambda I'd say.

Brian-Williams:
Thanks for the review!

Removed `require_exception`, which allowed us to inherit from `retry_if_exception` as suggested.

Removed `_no_failure` fnc as it no longer has a use case if `retry_if_not_exception_message` is the only `__call__` in this PR; `retry_if_exception_message` can simply inherit that magic method.

Fixed a bug in the assumption of the errormessage retrieval and wrote an additional unit test to handle the positive case of `retry_if_not_exception_message`.

Rebased on jd/master to a single commit.

Issue #134, opened 2018-08-08
Add min wait for exponential backoff


MikeWooster:
I have just came across the use case where I need to add a min wait with an exponential backoff.  We already have a `max` argument, would it be possible to add `min`?

PR #135, opened 2018-08-08
Added min to exponential backoff.

MikeWooster:
see: https://github.com/jd/tenacity/issues/134

### DIFF
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -143,9 +143,9 @@ services and other remote endpoints.
 
 .. testcode::
 
-    @retry(wait=wait_exponential(multiplier=1, max=10))
+    @retry(wait=wait_exponential(multiplier=1, min=4, max=10))
     def wait_exponential_1():
-        print("Wait 2^x * 1 second between each retry, up to 10 seconds, then 10 seconds afterwards")
+        print("Wait 2^x * 1 second between each retry starting with 4 seconds, then up to 10 seconds, then 10 seconds afterwards")
         raise Exception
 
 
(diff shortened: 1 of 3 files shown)


tenacity/_utils.py
```diff
@@ -26,7 +26,7 @@
     MAX_WAIT = sys.maxint / 2
 except AttributeError:
     MAX_WAIT = 1073741823
-
+MIN_WAIT = 0
```
jd: Just go ahead and use 0 directly. The `MAX_WAIT` here is just because you can't be sure what the max int might be.

tenacity/_utils.py
```diff
@@ -27,7 +27,6 @@
 except AttributeError:
     MAX_WAIT = 1073741823
 
-
```
jd: :(

MikeWooster:
@jd I have addressed all your comments.

doc/source/index.rst
```diff
@@ -143,9 +143,9 @@ services and other remote endpoints.
 
 .. testcode::
 
-    @retry(wait=wait_exponential(multiplier=1, max=10))
+    @retry(wait=wait_exponential(multiplier=1, min=4, max=10))
     def wait_exponential_1():
-        print("Wait 2^x * 1 second between each retry, up to 10 seconds, then 10 seconds afterwards")
+        print("Wait 2^x * 1 second between each retry starting with 4 seconds, then up to 10 seconds, then 10 seconds afterwards")
```
jd: "starting" might be confusing in the sense that 4 might be used several times if it's > than the computed value

tenacity/wait.py
```diff
@@ -158,7 +159,7 @@ def __call__(self, retry_state):
             result = self.multiplier * exp
         except OverflowError:
             return self.max
-        return max(0, min(result, self.max))
+        return max(max(0, self.min), min(result, self.max))
```
jd: i don't think the `max(0, self.min)` is useful, i'd just go on `self.min`.

PR #143, opened 2018-08-22
Remove extra increments of RetryCallState.attempt_number

WouldYouKindly:
Hello!

Both `AsyncRetrying.call` and `TornadoRetrying.call` incremented `RetryCallState.attempt_number` twice, which led to the following behavior:

```python
@retry(stop=stop_after_attempt(5), before=before_log(logger, logging.ERROR), reraise=True)
async def f():
    async with aiohttp.ClientSession() as s:
        await s.request('get', 'goo.coom')

asyncio.run(f())

...

Starting call to '__main__.f', this is the 1st time calling it.
Starting call to '__main__.f', this is the 3rd time calling it.
Starting call to '__main__.f', this is the 5th time calling it.
...
```

Seems like a bug.

### DIFF
--- a/tenacity/_asyncio.py
+++ b/tenacity/_asyncio.py
@@ -42,7 +42,6 @@ class AsyncRetrying(BaseRetrying):
         while True:
             do = self.iter(retry_state=retry_state)
             if isinstance(do, DoAttempt):
-                retry_state.attempt_number += 1
                 try:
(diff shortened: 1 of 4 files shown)


jd:
@WouldYouKindly squash your commits?

(ha! I can make a sentence with you nick :D)

WouldYouKindly:
@jd Please take a look (:

PR #168, opened 2019-04-19
Stop: Add stop_after_exception_count.

kobybum:
Add the ability to stop after a specific exceptions occurred X times.

### DIFF
--- a/tenacity/__init__.py
+++ b/tenacity/__init__.py
@@ -56,6 +56,7 @@ from .nap import sleep_using_event  # noqa
 # Import all built-in stop strategies for easier usage.
 from .stop import stop_after_attempt  # noqa
 from .stop import stop_after_delay  # noqa
+from .stop import stop_after_exception_count  # noqa
 from .stop import stop_all  # noqa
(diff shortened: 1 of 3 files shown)


tenacity/__init__.py
```diff
@@ -412,6 +414,8 @@ def __init__(self, retry_object, fn, args, kwargs):
 
         #: The number of the current attempt
         self.attempt_number = 1
+        #: Tracks exceptions raised by the method, by type
+        self.exception_counts = Counter()
```
jd: There should not be any need to store the number of exceptions here. The `stop_after_exception_count` class can store them AFAIU.

kobybum: You're right, I thought it would keep state between calls, updated

kobybum:
Nevermind, it looks like the state is kept between invocations, resulting in unwanted behavior.
This is the reason I opened the PR, since the handler doesn't work without inside modification.
I'll upload a failing test, let me know what you think, but this prevents us from using tenacity in a few places

jd:
Did you give up on this?

Issue #175, opened 2019-06-17
wait_exponential has unexpected first delay


cyounkins:
Hi there! Thank you for `tenacity`!

The first delay for `wait_exponential` is `2 * multiplier`, and to me that is unexpected. In version `5.0.4`:

```
In [14]: t0 = time.time()
    ...: t_last = t0
    ...: @retry(wait=wait_exponential(multiplier=0.2), stop=stop_after_attempt(5))
    ...: def wait():
    ...:     current_time = time.time()
    ...:     global t_last
    ...:     print("%.2f - %.2f" % (current_time - t0, current_time - t_last))
    ...:     t_last = current_time
    ...:     raise Exception
    ...: print("total - delta")
    ...: wait()
total - delta
0.00 - 0.00
0.40 - 0.40
1.21 - 0.80
2.81 - 1.60
6.01 - 3.20
```

To me it would make more sense for the first delay to be `1 * multiplier`. Others may expect the first retry to occur immediately with no delay, or maybe they want the sequence to be 0.1, 0.2, 0.4, 0.8, etc. which doesn't seem possible with just `wait_exponential`.

Can this be considered a bug? Is this unexpected to others, or do others want to be able to create the sequence above?

jd:
The first delay is like the rest of the, it's `(base_exp ** retry_attempt_number) * multiplier`, which is `2 * multiplier` if your base is `2` indeed.

So the default is: `2, 4, 8, 16`… which sounds good to me. Then it's up to you to change the `base` or the multiplier to tweak it`.

Unless I miss something?

cyounkins:
Yeah that makes sense. I think the reason I got confused is that when I think of the sequence `2, 4, 8, 16...` I think of the _multiplier_ as `2` where here it is the `base`.

So you can "back up" the sequence by dividing the `multiplier` by the `base_exp` or a power of `base_exp`. So to get 0.1, 0.2, 0.4, 0.8...

```
In [16]: t0 = time.time()
    ...: t_last = t0
    ...: @retry(wait=wait_exponential(multiplier=0.05), stop=stop_after_attempt(5))
    ...: def wait():
    ...:     current_time = time.time()
    ...:     global t_last
    ...:     print("%.2f - %.2f" % (current_time - t0, current_time - t_last))
    ...:     t_last = current_time
    ...:     raise Exception
    ...: print("total - delta")
    ...: wait()
    ...:
total - delta
0.00 - 0.00
0.10 - 0.10
0.31 - 0.20
[...]

PR #176, opened 2019-06-19
Reduce wait_exponential power by 1 so first wait is equal to multiplier

cyounkins:
Addresses #175 

The only apprehension I have about this is that it changes everyone's timing without them realizing.


jd:
Yeah, I agree with that, but I don't think it's a _huge_ issue. I need to fix the CI it seems.

jd:
@cyounkins sorry, I merged this by mistake while fighting with the CI. This needs to be reopened and the tests to be updated!

PR #191, opened 2019-08-27
Retry code block with context manager

bersace:
Should implement #48 

This PR implements only synchronous retrying for now. That should be a good starting point for this use case.

tenacity API is well designed, it's super easy to implement this with consistent behaviour !


bersace:
@jd :up: any idea on this ?

bersace:
@jd, mergify is awesome :-)

PR #198, opened 2019-10-30
Ensure that the _asyncio.py module can be compiled, even on Python 2

asqui:
Use the same pattern of conditional importing of _compat.py as used in
__init__.py to ensure that _compat.py can be compiled on Python 2.

Fixes jd/tenacity#88

### DIFF
--- a/tenacity/_asyncio.py
+++ b/tenacity/_asyncio.py
@@ -16,7 +16,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import asyncio
+try:
+    import asyncio
+except ImportError:
+    asyncio = None
+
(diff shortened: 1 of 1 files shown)


jd:
That might only work until we start using `await` or the like but let's do it ;)

asqui:
Cool, hopefully by then everyone will be on Python 3, right? :-)

Thanks for the super-speedy merge! Any plans for the next versioned release?

(No rush, just wondering -- I've already cherry-picked this change on top of v5.1.1, which we are using internally, so not having this PR in an official release has zero practical impact for me.)

jd:
Just pushed 5.1.5 with it.

Issue #197, opened 2019-10-30
Missing kwargs in retry helper in __init__.py


thehesiod:
on this section:
```python
    if len(dargs) == 1 and callable(dargs[0]):
        return retry()(dargs[0])
```
it should be:
```python
    if len(dargs) == 1 and callable(dargs[0]):
        return retry(**dkw)(dargs[0])
```

jd:
Would you mind sending a pull request with a unit test?

jd:
Not a bug 🤷‍♂

thehesiod:
this is a bug, you're dropping the `dkw` here: https://github.com/jd/tenacity/blob/master/tenacity/__init__.py#L96

jd:
Did you read the discussion on #199 @thehesiod?

thehesiod:
what about it?  it's a bad test, you're obviously dropping the kwargs on the floor so this is most definitely a bug.

PR #199, opened 2019-10-30
Propagate kwargs in retry helper (#197)

kestein:


tenacity/tests/test_tenacity.py
```diff
@@ -629,6 +629,17 @@ def test_retry_if_exception_message_negative_too_many_inputs(self):
             tenacity.retry_if_exception_message(
                 message="negative", match="negative")
 
+    def test_retry_kwargs_passed_properly(self):
+
+        def _kw_fun():
+            raise tenacity.TryAgain
+
+        g = retry(_kw_fun, stop=tenacity.stop_after_attempt(5), reraise=True)
```
jd: Hm, I'm not sure about that. It's not supposed to be used like that. It's supposedly a decorator.

You're supposed to do this here:

```python
g = retry(stop=tenacity.stop_after_attempt(5), reraise=True)(_kw_fun)
```

kestein: The syntax seems to support both but I swapped the test to be consistent 4d5b3fa

tenacity/tests/test_tenacity.py
```diff
@@ -629,6 +629,17 @@ def test_retry_if_exception_message_negative_too_many_inputs(self):
             tenacity.retry_if_exception_message(
                 message="negative", match="negative")
 
+    def test_retry_kwargs_passed_properly(self):
+
+        def _kw_fun():
+            raise tenacity.TryAgain
+
+        g = retry(stop=tenacity.stop_after_attempt(5), reraise=True)(_kw_fun)
```
jd: This works without your patch 🤷‍♂

PR #201, opened 2019-11-05
Support async keyword properly

### LINKED ISSUE #181: Asyncio: iscoroutinefunction does not work with tenacity
When decorating an `async` function with `@retry`, `asyncio.iscoroutinefunction(f)` unexpectedly returns `False`, because the wrapping function, in fact, is no coroutine.

This is a problem in some libraries that are using this way of determining if they are handling an async function or not (in my case: trafaret).

bersace:
This is easier since tenacity dropped support for 3.4.

Should fix #181


bersace:
@jd, ready for review.

Issue #210, opened 2020-01-08
Allow configurable defaults


afischer-opentext-com:
Hi, 

is there a way to re-configure the default configuration? I crawled the code base but failed to find something. 

In the code base I am maintaining the retry annotation is heavily used (more than 200 times). The vast majority of the calls looks equal, e.q.

```python
import logging

from tenacity import (before_sleep_log, retry, stop_after_attempt,
                      wait_exponential)

log = logging.getLogger(__name__)

@retry(before_sleep=before_sleep_log(log, logging.INFO), reraise=True, stop=stop_after_attempt(15), wait=wait_exponential(multiplier=1, min=1, max=10))
def something():
    pass
```

It would be great to have the ability to configure some kind of configuration group, something like the following as it help to reduce duplicate code.

```python
tenacity.config_group('foo', reraise=True, stop=stop_after_attempt(15), wait=wait_exponential(multiplier=1, min=1, max=10))

# and later

@retry(config_group='foo', before_sleep=before_sleep_log(log, logging.INFO))
def something():
    pass
```

markhu:
Could you kludge this with something like `@retry(eval(config_group['foo']))` ?

jd:
What's the problem with using something like:

```python
r = tenacity.Retrying(<your conf>)

r.wraps
def myfunc1():
    pass

r.wraps
def myfunc2():
    pass
```

?

afischer-opentext-com:
No problem! I opened the ticket asking for exactly something like this...
I was not aware about the functionality, it is not explicitly documented.
This also has the advantage that viewer objects are created.

PR #213, opened 2020-01-19
Implement Issue 210, configuration groups

slackfan:
Hi,

This PR implements a proposal for #210. It adds the necessary code, tests and documentation. 

The goal of the PR is to reduce the amount of duplicating configuration code the way it adds an optional method and and argument `configuration_group` which allows to configure and use configuration. 

A configuration group can just be used but also updated and oversteered when needed.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,4 +1,0 @@
-{
-  "python.testing.pytestEnabled": true,
-  "python.pythonPath": "/usr/bin/python3"
-}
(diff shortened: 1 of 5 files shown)


jd:
I'm not sure I see the whole point of this when you can just reuse your `Retrying` object for different functions?

slackfan:
I may not know something but how could I reuse a Retrying object when using the annotation?

slackfan:
Closing as the suggested solution is good enough.

Issue #204, opened 2019-11-15
DeprecationWarning re: async def in Python 3.8


hjwp:
silly little thing really.  Not sure if 3.8 is officially supported yet, if i get a mo i'll open a PR.

```
.venv/lib/python3.8/site-packages/tenacity/_asyncio.py:48
  /home/harry/workspace/cosmicpython/book/code/.venv/lib/python3.8/site-packages/tenacity/_asyncio.py:48: 
    DeprecationWarning: "@coroutine" decorator is deprecated since Python 3.8, use "async def" instead
    def call(self, fn, *args, **kwargs):

-- Docs: https://docs.pytest.org/en/latest/warnings.html
```

jd:
Which version of tenacity is this?

jd:
3.8 is supported

hjwp:
fair enough, just assumed it wasnt bc it wasn't in `tox.ini`.

PR #215, opened 2020-02-22
Use PEP 492 async/await syntax

andersea:
Hi, I just saw there was a larger PR with most of these changes.

This is just isolated modern async/await syntax.

I removed some code. This code doesn't bother with applying magic flags to the wrapped function to make it look like it is a coroutine. In my view, why bother with that check? The user should know what he is doing, so it should just work. 

I also tried to minimize asyncio dependancy to just tests and importing asyncio.sleep.

### DIFF
--- a/tenacity/__init__.py
+++ b/tenacity/__init__.py
@@ -18,9 +18,9 @@
 # limitations under the License.
 
 try:
-    import asyncio
+    from inspect import iscoroutinefunction
 except ImportError:
-    asyncio = None
+    iscoroutinefunction = None
 
 try:
     import tornado
@@ -96,7 +96,7 @@ def retry(*dargs, **dkw):
         return retry()(dargs[0])
     else:
         def wrap(f):
-            if asyncio and asyncio.iscoroutinefunction(f):
+            if iscoroutinefunction is not None and iscoroutinefunction(f):
                 r = AsyncRetrying(*dargs, **dkw)
             elif tornado and hasattr(tornado.gen, 'is_coroutine_function') \
                     and tornado.gen.is_coroutine_function(f):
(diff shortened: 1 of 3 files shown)


jd:
LGTM except one failure on py35 it seems.

andersea:
I am taking a look at the 3.5 build failure.

andersea:
Hmm.. I set up a VM in AWS with python3.5 and ran the tests. Seems to work fine there?!

```
(.venv) ubuntu@ip-172-30-0-191:~/tenacity$ pytest tenacity/tests/test_asyncio.py
==================================================================================== test session starts ====================================================================================
platform linux -- Python 3.5.9, pytest-5.3.5, py-1.8.1, pluggy-0.13.1
rootdir: /home/ubuntu/tenacity, inifile: setup.cfg
collected 4 items

tenacity/tests/test_asyncio.py ....                                                                                                                                                   [100%]

===================================================================================== 4 passed in 0.08s =====================================================================================
(.venv) ubuntu@ip-172-30-0-191:~/tenacity$
```

andersea:
I narrowed down the culprit. It is because of this change:

```
        future1 = asyncio.ensure_future(
            _retryable_coroutine.retry_with(after=after)(thing1))
        future2 = asyncio.ensure_future(
            _retryable_coroutine.retry_with(after=after)(thing2))

        yield from asyncio.gather(future1, future2)
```
Which I changed to -

```
        await asyncio.gather(
            _retryable_coroutine.retry_with(after=after)(thing1),
            _retryable_coroutine.retry_with(after=after)(thing2))
```

I dont understand the test. What is it doing?

andersea:
Right. So I changed the test slightly. It now checks if the set of even and odd members of the list are the same object, but doesn't expect thing1 to come before thing2. So it still checks that they are interleaved, but not in which order.

andersea:
Fixes #204

PR #217, opened 2020-03-18
Syntax highlighting in README.rst

batisteo:


batisteo:
Not working as expected.  
Neverwind :blush:

PR #218, opened 2020-03-22
Make autodoc to generate docs for retry_always()/retry_never()

sir-sigurd:
Related docs: https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#directive-autoattribute

### DIFF
--- a/tenacity/retry.py
+++ b/tenacity/retry.py
@@ -38,22 +38,20 @@ class retry_base(object):
 
 
 class _retry_never(retry_base):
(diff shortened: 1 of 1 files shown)


jd (commented):
Thank you! I'm a bit confused about why this is needed?
We're losing docstrings so I'd think it's a regression?

sir-sigurd:
> I'm a bit confused about why this is needed?

It's needed for retry_always()/retry_never() to be in API reference, because currently they're missing.

PR #222, opened 2020-04-03
AsyncRetrying can use an async iterator

willemt:
**Note:**
This feature doesn't support py35 because async generators were added in py36: https://www.python.org/dev/peps/pep-0525/

### DIFF
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -620,6 +620,20 @@ context. The trick is to combine a for loop and a context manager.
 You can configure every details of retry policy by configuring the Retrying
 object.
 
+With async code you can use AsyncRetrying.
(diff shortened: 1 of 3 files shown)


jd:
Could you add some tests for this?

We should make this work for Python 3.5 somehow if we can. :( or at least not make it fail completely.

willemt:
hi @jd I've made it Python3.5 friendly. Also tests have been included

willemt:
hi @jd just wanted to check if you needed anything else for this PR to be merged in? thanks

achimnol:
I just discovered this library and found that the async-version of block-style retrying does not work....
Voting for this PR! 👍

PR #231, opened 2020-05-03
Do not pretend to contain type annotations

### LINKED ISSUE #230: Version 6.2.0 type hinting breaks mypy with --strict
With the following simple program, running `mypy --strict` claims the methods are not part of the tenacity module. Amusingly, mypy suggests other names in the tenacity module, and if you add more code using those names, mypy complains about them also and suggests the original names it complained about. Adding `# type: ignore` comments does silence the errors, but that's not ideal.

```python
import tenacity


@tenacity.retry(
    retry=tenacity.retry_if_exception_type(RuntimeError),
    wait=tenacity.wait_exponential(exp_base=1.2, multiplier=1 / 1.2),
    stop=tenacity.stop_after_attempt(10),
    reraise=True,
)
def keep_trying() -> None:
    raise RuntimeError("Oops")


keep_trying()
```

```console
$ mypy --strict main.py
main.py:5: error: Module has no attribute "retry_if_exception_type"; maybe "retry_if_exception", "retry_unless_exception_type", or "retry_if_exception_message"?
main.py:6: error: Module has no attribute "wait_exponential"; maybe "wait_random_exponential"?
main.py:7: error: Module has no attribute "stop_after_attempt"
Found 3 errors in 1 file (checked 1 source file)
```

JenSte:
According to PEP 561, a file named py.typed in a package implies that the
code contains type hints. This is currently not the case, so remove it.

Resolves #230

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -28,9 +28,6 @@ install_requires =
     typing>=3.7.4.1;python_version=='2.7'
 packages = tenacity
 
-[options.package_data]
-tenacity = py.typed
-
 [options.extras_require]
 doc =
     reno


jd:
It has some hints though. Is removing this file disabling the existing hints?

JenSte:
It was always my understanding that adding a "py.typed" file to a package means that it is fully covered with type hints. That understanding probably came from a few modules that added full type hints and only then the py.typed file, and then allowed me to remove the "# type: ignore" comment after the import of the module in my code.

But I have to admit that I can not find this strict requirement in PEP 484 or 561, they talk about "supporting typing" without actually saying if that means partial type hints, type hints for all functions that are exposed to the outside, or full type hints for all functions. PEP 561 has a section on partial typing that states "If a stub package is partial it MUST include partial\n in a top level py.typed file.". However we here have the problem with inline type hints and not with stub packages, and adding that string to py.typed did not work, `mypy --strict` still complains about `Call to untyped function in typed context`. :-/

> Is removing this file disabling the existing hints?

I removed the py.typed and then typechecked the `do_something_unreliable()` example from the readme, and unfortunately `mypy --strict` started complaining after the file was removed.

jd:
My understanding is that having this file means the package contains the hints, if there exists, rather than adding them in typeshed.
I'm more than happy to have type hints in this package.

I'm closing this, but feel free to reopen if I'm totally wrong.

PR #253, opened 2020-09-14
Make AsyncRetrying callable (as Retrying class)

ivanprado:
Not entirely sure, but I would say that this line should be added so that it follows the same structure that `Retrying` class.

### DIFF
--- a/tenacity/__init__.py
+++ b/tenacity/__init__.py
@@ -16,7 +16,6 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-from abc import abstractmethod, ABCMeta
 
 try:
     from inspect import iscoroutinefunction
@@ -31,6 +30,7 @@ except ImportError:
 import sys
 import threading
 import typing as t
+from abc import ABCMeta, abstractmethod
 from concurrent import futures
 
 import six
--- a/tenacity/__init__.py
+++ b/tenacity/__init__.py
@@ -16,6 +16,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+from abc import abstractmethod, ABC
 
 try:
     from inspect import iscoroutinefunction
@@ -208,7 +209,7 @@ class AttemptManager(object):
             self.retry_state.set_result(None)
 
 
-class BaseRetrying(object):
+class BaseRetrying(ABC):
 
     def __init__(self,
                  sleep=sleep,
@@ -326,7 +327,7 @@ class BaseRetrying(object):
         """
         @_utils.wraps(f)
         def wrapped_f(*args, **kw):
-            return self.call(f, *args, **kw)
+            return self(f, *args, **kw)
 
         def retry_with(*args, **kwargs):
             return self.copy(*args, **kwargs).wraps(f)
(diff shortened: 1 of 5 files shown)


ivanprado:
To give you more context. I'm using `AsyncRetrying` as:
```py
async def task():
  pass

retrying = AsyncRetrying(...)
await retrying.call(task)
```

But the sync version would be different (according to the [documentation](https://tenacity.readthedocs.io/en/latest/index.html#retrying-code-block)):

```py
def task():
  pass

retrying = Retrying(...)
retrying(task)
```

Note that `call` invokation is not required in the later. So, at first sight, I would say that this PR makes sense. 

Let me know if you think this makes sense. If this is the case I'll try to implement a unit test.

jd:
Yeah makes sense to me. I think we assumed it was available on `BaseRetrying`. Does it make sense to move it there?

ivanprado:
>  Does it make sense to move it there?

@jd I've been thinking about it. My proposal would be to just directly override the `__call__` in both `Retrying` and `AsyncRetrying`. So instead: 

```py
class AsyncRetrying(BaseRetrying):

  async def call(...):
     pass

  __call__ = call
```

We would have: 
```py
class AsyncRetrying(BaseRetrying):
  async def call(...):
    pass
```

The same for `Retrying` but without the `async`. 

This would also require changing the line https://github.com/jd/tenacity/blob/master/tenacity/__init__.py#L329 for:
`return self(f, *args, **kw)` 
but I think this is Ok. 

What do you think?

ivanprado:
@jd I've been updating the proposal. 

* Now all classes extending `BaseRetrying` are forced to implement `__call__`. 
* `Retrying`, `AsyncRetrying` and `TornadoRetying` are updated accordingly. 
* All unit tests are updated to use `__call__` instead of `call`. 
* A new unit test was included to test this interface on `AsyncRetrying`. 

Let me know what do you think.

tenacity/__init__.py
```diff
@@ -396,11 +398,15 @@ def __iter__(self):
...
+    @abstractmethod
+    def __call__(self, *args, **kwargs):
+        pass
+
 
 class Retrying(BaseRetrying):
     """Retrying controller."""
 
-    def call(self, fn, *args, **kwargs):
+    def __call__(self, fn, *args, **kwargs):
```
jd, line 409: This rename might break the API, right?

ivanprado: Well, I didn't find in any part of the documentation neither the API docs (see here https://tenacity.readthedocs.io/en/latest/api.html?highlight=Retrying#tenacity.Retrying) where the method `call` is mentioned. 

But nothing prevents that somebody noticed that the method exists and is public, so somebody could be using it. In fact, in my code, I'm using it as a workaround. 

I don't know which is the tenacity policy regarding that. 

I could do something like the following for `Retrying` class only:

```py
  def call(self, *args, **kwargs):
    """Deprecated. Use ``__call__`` instead"""
    warnings.warn("'call' function is deprecated. Use '__call__' instead")
    self.__call__(*args, **kwargs)
```

This would keep backwards compatibility. @jd what do you think?

jd: Works for me 👍

ivanprado: @jd implemented here: https://github.com/jd/tenacity/pull/253/commits/a5830a952a93db79acc7a4394f881bec12f2e316 and here: https://github.com/jd/tenacity/pull/253/commits/5245c747e7e865a1ecc3e8319193e417dc7da4d1

Let me know what do you think.

jd (approved):

ivanprado:
Thank you @jd for the review.

tenacity/__init__.py
asqui: A little late to the party given this PR is merged, but should this raise a `DeprecationWarning` (rather than the default `UserWarning`) to match other parts of tenacity?

PR #258, opened 2020-10-20
Fix a number of issues with pr253

asqui:
1. Add test coverage for `Retrying.__call__()` interface to demonstrate a number of issues with the changes from #253 
1. Fix the issues with the backward compatibility `Retrynig.call()` method:-
- It calls `self.__call__(self, ...)` i.e. passing itself, the Retrying instance, as the function to call. Now that Retrying is callable, this sneaks in and leads to reentrant invocation of the same Retrying object (is that even supported?) and, ultimately nested retrying. Uber-badness; not caught by any existing tests! :-o
- It doesn't return the result.
- It raises a `UserWarning` rather than `DeprecationWarning` type (which is what other parts of the code use for backward compatibility shims)

This PR contains a number of commits that can and probably should be squashed into a single commit after review. Let me know if you want me to do this (I don't know how easy GitHub makes it for you to do this at merge-time, whilst letting you compose a sensible commit message).

### DIFF
--- a/tenacity/__init__.py
+++ b/tenacity/__init__.py
@@ -431,7 +431,7 @@ class Retrying(BaseRetrying):
     def call(self, *args, **kwargs):
         """Use ``__call__`` instead because this method is deprecated."""
         warnings.warn("'Retrying.call()' method is deprecated. " +
-                      "Use 'Retrying.__call__()' instead")
+                      "Use 'Retrying.__call__()' instead", DeprecationWarning)
         return self.__call__(*args, **kwargs)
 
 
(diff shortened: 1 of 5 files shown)


tenacity/__init__.py
```diff
@@ -431,8 +431,8 @@ def __call__(self, fn, *args, **kwargs):
     def call(self, *args, **kwargs):
         """Use ``__call__`` instead because this method is deprecated."""
         warnings.warn("'Retrying.call()' method is deprecated. " +
-                      "Use 'Retrying.__call__()' instead")
-        self.__call__(self, *args, **kwargs)
+                      "Use 'Retrying.__call__()' instead", DeprecationWarning)
+        return self.__call__(*args, **kwargs)
```
ivanprado, line 435: Ops, what a big miss. Thank you for finding this!

jd (approved):

Issue #224, opened 2020-04-09
Regression: Asyncio: iscoroutinefunction does not work with tenacity


sbrandtb:
#181 returned. It was fixed in 3bae79e71519ed1a34cea53dcf98903ef02eb6c7, but apparently broke again in 66de0110844a742b820ab6c1f29edb19ed2e5908.

```py
from inspect import iscoroutinefunction

from tenacity import retry


async def a():
    pass

print(iscoroutinefunction(a))

b = retry(a)

print(iscoroutinefunction(b))
```

```
True
False
```

Again, this breaks everything that depends on the coroutine test :slightly_frowning_face:

jd:
I guess we missed a test for this. Feel free to send a PR for it!

NatureGeorge:
It seems that this issue has not been perfectly solved.

```bash
python -m pip install git+https://github.com/jd/tenacity.git@be9bccbb731d28ab0bcd8a4ee5acae70988bf593
```

```python
from inspect import iscoroutinefunction as inspect_iscoroutinefunction
from asyncio import iscoroutinefunction as asyncio_iscoroutinefunction

from tenacity import retry


async def a():
    pass

print(inspect_iscoroutinefunction(a), asyncio_iscoroutinefunction(a))

b = retry(a)

print(inspect_iscoroutinefunction(b), asyncio_iscoroutinefunction(b))
```

```bash
True True
False True
```

PR #265, opened 2020-12-14
Fix iscoroutinefunction check (for both inspect & asyncio)

NatureGeorge:
#263 did not fully solve #224.

## current behavior

```bash
python -m pip install git+https://github.com/jd/tenacity.git@be9bccbb731d28ab0bcd8a4ee5acae70988bf593
```

```python
from inspect import iscoroutinefunction as inspect_iscoroutinefunction
from asyncio import iscoroutinefunction as asyncio_iscoroutinefunction

from tenacity import retry


async def a():
    pass

print(inspect_iscoroutinefunction(a), asyncio_iscoroutinefunction(a))

b = retry(a)

print(inspect_iscoroutinefunction(b), asyncio_iscoroutinefunction(b))
```

```bash
True True
False True
```

## How to solve

https://github.com/jd/tenacity/blob/be9bccbb731d28ab0bcd8a4ee5acae70988bf593/tenacity/_asyncio.py#L78

The reason that it passes `asyncio.iscoroutinefunction`  is:

> https://docs.python.org/3/library/asyncio-task.html#asyncio.iscoroutinefunction

> asyncio.iscoroutinefunction(func)
> Return True if func is a coroutine function.
> This method is different from inspect.iscoroutinefunction() because it returns True for generator-based coroutine functions decorated with @coroutine.

But it can not pass `inspect.iscoroutinefunction` since the returned `fn` is not an `async def` function.

Through explicitly defining an `async` function (wrapper) in https://github.com/jd/tenacity/blob/be9bccbb731d28ab0bcd8a4ee5acae70988bf593/tenacity/_asyncio.py#L75

one can make it pass both `inspect.iscoroutinefunction` and `asyncio.iscoroutinefunction`.

### DIFF
--- a/tenacity/_asyncio.py
+++ b/tenacity/_asyncio.py
@@ -74,10 +74,10 @@ class AsyncRetrying(BaseRetrying):
     def wraps(self, fn):
         fn = super().wraps(fn)
         # Ensure wrapper is recognized as a coroutine function.
-        
+
         async def async_wrapped(*args, **kwargs):
             return await fn(*args, **kwargs)
-        
+
         # Preserve attributes
         async_wrapped.retry = fn.retry
         async_wrapped.retry_with = fn.retry_with
--- a/tenacity/_asyncio.py
+++ b/tenacity/_asyncio.py
@@ -15,7 +15,6 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-import asyncio
 import sys
 from asyncio import sleep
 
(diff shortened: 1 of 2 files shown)


tenacity/_asyncio.py
```diff
@@ -75,5 +74,12 @@ async def __anext__(self):
...
         # Ensure wrapper is recognized as a coroutine function.
-        fn._is_coroutine = asyncio.coroutines._is_coroutine
-        return fn
+
+        async def async_wrapped(*args, **kwargs):
+            return await fn(*args, **kwargs)
+
+        # Preserve attributes
+        async_wrapped.retry = fn.retry
+        async_wrapped.retry_with = fn.retry_with
```
jd, line 83: maybe add tests on this too just to be sure?

jd (changes requested):

tenacity/_asyncio.py
NatureGeorge: I think this test is enough for ensuring the attributes:

https://github.com/jd/tenacity/blob/be9bccbb731d28ab0bcd8a4ee5acae70988bf593/tenacity/tests/test_asyncio.py#L94-L96

Still, perhaps it would be safer to have a stand-alone test for this.

jd (approved):

PR #266, opened 2020-12-15
Add testing for Python 3.9 and PyPy3

jdufresne:
Trove classifiers were added for both to document support.

Python 3.9 was released on October 5th, 2020.


jd (changes requested):
I don't think we're running pypy in the ci, right?

Issue #269, opened 2020-12-25
retry() hangs indefinitely (and silently) if a retry_base is passed positionally


SnoopJ:
If a user decorates a function with `@retry(inst)` for a `retry_base` instance `inst` for any of the builtin strategies as a positional argument—an easy misspelling of `@retry(retry=inst)`)—`tenacity` will hang indefinitely as the resulting decoration attempts to call the _strategy_ with the function as an argument, which will generally be an error.

Below is a program that recreates this failure for all builtin strategies.

```python
from multiprocessing import Pool, Process
import sys

from tenacity import retry
from tenacity.retry import *


def make_class(strat):
    """
    This sample will hang indefinitely while defining the class Bar

    retry() thinks the strategy is what I want to retry,the unrolled
    equivalent is:

          def retryable(self, param):
              pass
          retryable = retry(strat)(retryable)

    so the thing being retried is strat(retryable) which is very unlikely
    to be a valid call
    """
    class Bar:
        print("Defining Bar...")

        @retry(strat)  # oops! I forgot to type retry=
        def retryable(self, param):
            pass

        print("Done defining Bar")  # we never get here

    return Bar
def make_func(strat):                                                                                                                  [1/1317]
    """
    This sample will hang indefinitely while defining the function task()
    """
    @retry(strat)  # oops! I forgot to type retry=
    def task(x):
        print("Running task")  # will never print
        pass

    return task

strats = {
    "retry_if_exception": retry_if_exception(lambda e: True),
    "retry_if_exception_type": retry_if_exception_type(TimeoutError),
    "retry_unless_exception_type": retry_unless_exception_type(RuntimeError),
    "retry_if_result": retry_if_result(lambda e: True),
    "retry_if_not_result": retry_if_not_result(lambda e: False),
    "retry_if_exception_message": retry_if_exception_message(match=".*"),
    "retry_if_not_exception_message": retry_if_not_exception_message(match=".*"),
    "retry_any": retry_any(retry_if_result(lambda x: lambda y: True),),
    "retry_all": retry_all(retry_if_result(lambda x: lambda y: True),),
}


TIMEOUT = 5

if __name__ == "__main__":
    for name,strat in strats.items():
        for f in (make_class, make_func):
            proc = Process(target=f, args=(strat,))
            print(f"Calling {f.__name__} with arg {name} with a timeout of {TIMEOUT} seconds")
            proc.start()
            proc.join(TIMEOUT)  # give the process a few seconds
[...]

PR #270, opened 2020-12-25
Warn when calling retry() on retry_base (#269)

SnoopJeDi:
This PR adds a warning when a `retry_base` instance is passed to `tenacity.retry` as a positional argument (see #269)

It might be reasonable to make this warning an error, or a specific class of warning that can be converted to an error by users who wish to do so, but emitting some kind of feedback in this probably-wrong case can help avoid unintuitive hanging of programs where this flaw presents.


jd:
LGTM, CI fails though.

SnoopJeDi:
Most of the CI failures were #276 but one was a line-width problem on the warning. Sorry about that!

jd (approved):

Issue #284, opened 2021-03-10
Changelog entries missing for > 6.3.0


hynek:
There seems to be a bit of inconsistency between:

- https://pypi.org/project/tenacity/#history (latest release 7.0.0)
- https://tenacity.readthedocs.io/en/latest/changelog.html (6.3.0)
- https://github.com/jd/tenacity/releases (5.1.4)

Any chance we could learn what the supposedly breaking changes were? 😇

jd:
The idea was to use `reno` to manage release notes, except that this has been missed multiple times.

I've added a check for this now in #285 so we're sure it stops happening.

If anyone is brave enough to write previous changelog entries, we can try to put them somewhere I guess.

pheanex:
This is the list of commits since 6.3.1:
* [Avoid including tests when packaging (#281)](https://github.com/jd/tenacity/commit/3b5c6bff5815fb86627b1fc693524f936058b61f)
* [Warn when calling retry() on retry_base (#269) (#270)](https://github.com/jd/tenacity/commit/3e2244535ccfbb6a4b7fdd77bfc9aa61a1302302)
* [Copy whole internal state when retry_with (#233) (#278)](https://github.com/jd/tenacity/commit/68faeb07c19a609b11e9e377ac210c266fa74435)
* [CI: Enable black (#277)](https://github.com/jd/tenacity/commit/fd7842774cc10d41e68e6bc64afe7d9f83221aa0)
* [Merge pull request #275 from jd/py39](https://github.com/jd/tenacity/commit/369ac3d2ef7aa7946ac767f8e055f841ca4ac7d2)
* [Merge branch 'master' into py39](https://github.com/jd/tenacity/commit/693ab1588fb44c88b5f43d652d5d82f7a4a2e07b)
* [Refactor: Remove old compat support code (#274)](https://github.com/jd/tenacity/commit/250b27e47712e42e096a68a3c3518da8de0e90ba)
* [Chore: Add support for Python 3.9](https://github.com/jd/tenacity/commit/5e8e97aa325dd15d0cf8bd2ead08f9b288f09b31)
* [Pull Request for fix-ci (#276)](https://github.com/jd/tenacity/commit/2214708c7e95338bd64700fae399953364148e52)
* [Fix: Remove saythanks (dead)](https://github.com/jd/tenacity/commit/8d629023f8c95a729c915ff4c4bf415ad28f59f9)

@jd can you tell which one required a major version bump?

jd:
[...]

PR #285, opened 2021-03-11
ci(mergify): force release notes to be present

jd:
This should make sure we don't forget to put a release note before merging a
PR.

Related to #284


asqui:
How very meta... the bot is complaining that you didn't add a changelog entry for the change that mandates changelog entries. 😂

jd:
I guess it works… too well 😆

PR #292, opened 2021-04-15
Before sleep log stack info

### LINKED ISSUE #290: Allow stack_info argument in before_sleep_log strategy
#227 adds the `exc_info` parameter to the `before_sleep_log()` function.  This is nice, but the stack trace ends at the place where the exception is raised.  It may be useful to have the entire stack context as part of the log by allowing the stack_info argument to the logger.

See description in the logging documentation:

> The second optional keyword argument is stack_info, which defaults to False. If true, stack information is added to the logging message, including the actual logging call. Note that this is not the same stack information as that displayed through specifying exc_info: The former is stack frames from the bottom of the stack up to the logging call in the current thread, whereas the latter is information about stack frames which have been unwound, following an exception, while searching for exception handlers.

> You can specify stack_info independently of exc_info, e.g. to just show how you got to a certain point in your code, even when no exceptions were raised. The stack frames are printed following a header line which says:

`Stack (most recent call last):`

> This mimics the Traceback (most recent call last): which is used when displaying exception frames.

Note, I have this coded up.  Just let me know if this is okay and I can open a PR.

jpark712:
This change adds the `stack_info` parameter to the before_sleep_log() function and passes it to the logger function.  When set to True, logger will include a stack dump at the current location, which can be useful to pinpointing the exact location in code that triggered a retry.

Closes #290

### DIFF
--- a/releasenotes/notes/before_sleep_log-stack_info-ec404d38a82f4f9d.yaml
+++ b/releasenotes/notes/before_sleep_log-stack_info-ec404d38a82f4f9d.yaml
@@ -1,0 +1,3 @@
+---
+features:
+  - Add a ``stack_info`` option to the ``before_sleep_log()`` strategy.
--- a/tenacity/before_sleep.py
+++ b/tenacity/before_sleep.py
@@ -22,7 +22,7 @@ def before_sleep_nothing(retry_state):
     """Before call strategy that does nothing."""
 
 
-def before_sleep_log(logger, log_level, exc_info=False):
+def before_sleep_log(logger, log_level, exc_info=False, stack_info=False):
     """Before call strategy that logs to some logger the attempt."""
 
     def log_it(retry_state):
@@ -46,6 +46,7 @@ def before_sleep_log(logger, log_level, exc_info=False):
             verb,
             value,
             exc_info=local_exc_info,
+            stack_info=stack_info,
         )
 
     return log_it
--- a/tenacity/tests/test_tenacity.py
+++ b/tenacity/tests/test_tenacity.py
@@ -1174,7 +1174,7 @@ class TestBeforeAfterAttempts(unittest.TestCase):
         logger.addHandler(handler)
         try:
             _before_sleep = tenacity.before_sleep_log(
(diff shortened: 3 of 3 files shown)


jpark712:
Example of before_sleep_log called with exc_info=True, stack_info=False:

```
test_func called()
INFO:__main__:Retrying __main__.test_func in 1.0 seconds as it raised Exception: my error.
  File "C:\github\forks\tenacity\tenacity\__init__.py", line 433, in __call__
    result = fn(*args, **kwargs)
  File ".\ten.py", line 35, in test_func
    raise Exception("my error")
Exception: my error
```

Example of before_sleep_log called with exc_info=True, stack_info=True.  Notice we can now see that test_func was called from line 117 from the main() method whereas before, we only get the line that raised the exception.
```
test_func called()
INFO:__main__:Retrying __main__.test_func in 1.0 seconds as it raised Exception: my error.
  File "C:\github\forks\tenacity\tenacity\__init__.py", line 433, in __call__
    result = fn(*args, **kwargs)
  File ".\ten.py", line 35, in test_func
    raise Exception("my error")
Exception: my error
Stack (most recent call last):
  File ".\ten.py", line 117, in <module>
    main()
  File ".\ten.py", line 94, in main
    r.call(test_func, 1, 2)
  File "C:\github\forks\tenacity\tenacity\__init__.py", line 419, in call
    return self.__call__(*args, **kwargs)
  File "C:\github\forks\tenacity\tenacity\__init__.py", line 430, in __call__
    do = self.iter(retry_state=retry_state)
  File "C:\github\forks\tenacity\tenacity\__init__.py", line 391, in iter
    self.before_sleep(retry_state=retry_state)
  File "C:\github\forks\tenacity\tenacity\before_sleep.py", line 41, in log_it
    logger.log(
```

jd (approved):

jd:
Ah I think CI is broken again. Really need to move to GH Actions.

RoyalTS:
This would be super useful to have. Is there any chance of landing this soon?

jd:
Old and in conflict, closing.

Issue #291, opened 2021-04-14
Drop Python 2 support


and-semakin:
Python 2 has [reached its EOL](https://www.python.org/dev/peps/pep-0373/) about 1.5 years ago. Even AWS Lambda is going to [deprecate their Python 2 runtime](https://aws.amazon.com/blogs/compute/announcing-end-of-support-for-python-2-7-in-aws-lambda/) soon. Supporting this old version of Python prevents using modern language features, for example, PEP-484 inline type hints (not those comments).

You can go the same way as [guys from Pallets Project did](https://palletsprojects.com/blog/ending-python2-support/) and use the `python_requires` package metadata. It will allow you to release the next version without Python 2 support (let's say 6.0.0), but Python 2 users will still be able to download and use the latest available version with Python 2 support (let's say 5.1.5).

Is there any plan of dropping Python 2? What do you think?

and-semakin:
As we can see [here](https://pypistats.org/packages/tenacity) there are approximately 15% of installations still happening from Python 2:

![image](https://user-images.githubusercontent.com/9129071/114714839-eda42e00-9d4b-11eb-81de-5358fd2e544f.png)

and-semakin:
Also, [Python 3.5](https://github.com/jd/tenacity/blob/6855c55d9d33da3c62f12f6b198232e050b3988e/setup.cfg#L17) is deprecated as well.

jd:
Fine with me if you go with `python_requires` so older version are available for Python 2.
I'm fine dropping Python 3.5 too.

and-semakin:
Did some experiments and found that it's safe to drop Python 2 support right away, without needing an additional release with `python_requires` that will include `2.7`. Check it out: #293

PR #293, opened 2021-04-17
Drop support for deprecated Pythons

and-semakin:
This PR obsoletes Python 2.7 & 3.5, removes them from package metadata and CI. Also, it adds the `python_requires` metadata option which instructs `pip` running on older Python versions to find and install the latest suitable release.

See #291

### DIFF
--- a/releasenotes/notes/drop-deprecated-python-versions-69a05cb2e0f1034c.yaml
+++ b/releasenotes/notes/drop-deprecated-python-versions-69a05cb2e0f1034c.yaml
@@ -1,0 +1,4 @@
+---
+other:
+  - |
+    Drop support for deprecated Python versions (2.7 and 3.5)
--- a/setup.cfg
+++ b/setup.cfg
@@ -11,10 +11,8 @@ classifier =
     Intended Audience :: Developers
     License :: OSI Approved :: Apache Software License
     Programming Language :: Python
-    Programming Language :: Python :: 2
-    Programming Language :: Python :: 2.7
     Programming Language :: Python :: 3
-    Programming Language :: Python :: 3.5
+    Programming Language :: Python :: 3 :: Only
     Programming Language :: Python :: 3.6
     Programming Language :: Python :: 3.7
     Programming Language :: Python :: 3.8
@@ -24,9 +22,7 @@ classifier =
 [options]
 install_requires =
     six>=1.9.0
-    futures>=3.0;python_version=='2.7'
-    monotonic>=0.6;python_version=='2.7'
-    typing>=3.7.4.1;python_version=='2.7'
+python_requires = >=3.6
 packages = tenacity
 
 [options.packages.find]
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,5 @@
 [tox]
-envlist = py27, py35, py36, py37, py38, py39, pep8, pypy
+envlist = py36, py37, py38, py39, pep8, pypy3
 
 [testenv]
 usedevelop = True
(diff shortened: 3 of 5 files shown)


setup.cfg
```diff
@@ -11,10 +11,8 @@ classifier =
     Intended Audience :: Developers
     License :: OSI Approved :: Apache Software License
     Programming Language :: Python
-    Programming Language :: Python :: 2
-    Programming Language :: Python :: 2.7
     Programming Language :: Python :: 3
-    Programming Language :: Python :: 3.5
+    Programming Language :: Python :: 3 :: Only,
```
jd, line 15:
```suggestion
    Programming Language :: Python :: 3 :: Only
```

setup.cfg
```diff
@@ -24,9 +22,7 @@ classifier =
 [options]
 install_requires =
     six>=1.9.0
-    futures>=3.0;python_version=='2.7'
-    monotonic>=0.6;python_version=='2.7'
-    typing>=3.7.4.1;python_version=='2.7'
+python_requires = >=3.6, <4
```
jd, line 25: you sure it does not work on Python 4?

jd (changes requested):

setup.cfg
```diff
@@ -24,9 +22,7 @@ classifier =
 [options]
 install_requires =
     six>=1.9.0
```
Dunedan, line 24: I'd appreciate if the requirement on `six` would get dropped together with Python 2.7 support as well.

setup.cfg
penguinolog: Officially announced, that changes like 2 -> 3 is not expected in the future. If it will be changed - it will be reason to make such limitation

setup.cfg
and-semakin: Thanks for noticing! Fixed.

setup.cfg
and-semakin: That's what some popular projects do ([urllib3](https://github.com/urllib3/urllib3/blob/55465173b3eb233e14159908834c64367c28e203/setup.py#L109)) and what some package managers force you to do ([poetry](https://github.com/python-poetry/poetry/issues/1413)). I explored this topic a little and found that [people in PyPA prefer to not have an upper bound](https://discuss.python.org/t/use-of-less-than-next-major-version-e-g-4-in-python-requires-setup-py/1066/3). So, let it be this way. Removed `<4`.

setup.cfg
and-semakin: Yeah, it could be nice. Totally agree but I'd prefer to make it in a separate PR to keep this one focused. After removing Python 2 dependency we could have a bunch of other different improvements. Created an issue to not forget about it: https://github.com/jd/tenacity/issues/305

penguinolog:
Ooh, last update was so long time ago, that I was afraid that it abandoned.
You can borrow next changes from mine pr then: I made several things at
once: drop old + drop compatibility + fix broke mypy annotations (used to
check, that minimally not broken internals) and tested on python up to 3.10

and-semakin:
@homeworkprod updated the commit message, now it contains exact Python versions, thanks!

and-semakin:
@penguinolog, wow, you've done a huge amount of work! Great job! I borrowed a little from your PR. Thanks!

penguinolog (approved):

penguinolog:
I'll rebase mine changes after merge of this PR

releasenotes/notes/drop-deprecated-python-versions-69a05cb2e0f1034c.yaml
```diff
@@ -0,0 +1,4 @@
+---
+other:
+  - |
+    Drop support for deprecated Python version (2.7 and 3.5)
```
jd, line 4:
```suggestion
    Drop support for deprecated Python versions (2.7 and 3.5)
```

tox.ini
```diff
@@ -1,18 +1,17 @@
...
     .[doc]
     pytest
-    typeguard;python_version>='3.0'
+    typeguard
 commands =
-    py{27,py}: pytest --ignore='tenacity/tests/test_asyncio.py' {posargs}
-    py3{5,6,7,8,9}: pytest {posargs}
-    py3{5,6,7,8,9}: sphinx-build -a -E -W -b doctest doc/source doc/build
-    py3{5,6,7,8,9}: sphinx-build -a -E -W -b html doc/source doc/build
+    py3{6,7,8,9,py3}: pytest {posargs}
```
jd, line 12: This ends up being `py3py3` I don't think it'll work :)

jd (dismissed):

releasenotes/notes/drop-deprecated-python-versions-69a05cb2e0f1034c.yaml
and-semakin: Fixed, thanks!

tox.ini
and-semakin: Ah good catch, thanks! Fixed.

jd (approved):

and-semakin:
Thanks for your thorough review @jd!

PR #294, opened 2021-04-22
Make logger more compatible

shadowwalker:
Use `str.format` to complete format string internally, this helps `tenacity` to be more compatible with other loggers such as `loguru`.

Without this PR the logging shows `tenacity.before_sleep:log_it:41 - Retrying %s in %s seconds as it %s %s.` when passing in logger from loguru.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,5 @@ AUTHORS
 ChangeLog
 .eggs/
 doc/_build
-.venv/
 
 /.pytest_cache
--- a/.gitignore
+++ b/.gitignore
@@ -8,5 +8,6 @@ AUTHORS
 ChangeLog
 .eggs/
 doc/_build
+.venv/
 
 /.pytest_cache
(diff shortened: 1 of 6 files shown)


.gitignore
```diff
@@ -8,5 +8,6 @@ AUTHORS
 ChangeLog
 .eggs/
 doc/_build
+.venv/
```
jd, line 11:
```suggestion
```

tenacity/__init__.py
```diff
@@ -102,7 +102,7 @@ def retry(fn):
 @t.overload
 def retry(*dargs, **dkw):  # noqa
     # type: (...) -> t.Callable[[WrappedFn], WrappedFn]
-    """Type signature for the @retry() decorator constructor."""
+    """Type signature for the @retry() decorator constructor."""  # noqa: D402
```
jd, line 105: Sounds unrelated?

jd (dismissed):
Not a fan of this. Loguru not implementing the standard logging interface is their choice but making it paid by every library out there is not an option.

Since in that case, it's not too hard, I'd be ok with this. I'm just worried it'll be broken since there's no test and no easy way to add tests.

.gitignore
shadowwalker: Updated

tenacity/__init__.py
shadowwalker: Updated

jd:
@mergifyio update

jd (approved):

PR #295, opened 2021-04-25
Support Trio out-of-the-box

njsmith:
This PR makes `@retry` just work when running under Trio.

For now, I added `trio` to the test deps in tox.ini, so we can test
it. The code is also designed to handle things gracefully if trio
isn't installed, but currently there's not any automated test for
that... how would you like to handle this? Maybe add another
test environment without trio?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -64,6 +64,15 @@ jobs:
           command: |
             sudo pip install tox
             tox -e py39
+  py39-notrio:
+    docker:
+      - image: circleci/python:3.9
+    steps:
+      - checkout
+      - run:
+          command: |
+            sudo pip install tox
+            tox -e py39-notrio
   deploy:
     docker:
       - image: circleci/python:3.9
@@ -100,6 +109,7 @@ workflows:
       - py37
       - py38
       - py39
+      - py39-notrio
       - deploy:
           filters:
             tags:
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -546,28 +546,34 @@ With async code you can use AsyncRetrying.
 Async and retry
 ~~~~~~~~~~~~~~~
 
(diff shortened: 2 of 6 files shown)


jd (dismissed):
Yeah I think we should cover both. Just add a new tox scenario with and without trio :)

njsmith:
@jd How's this look?

jd (changes requested):
I think it's good though I'd prefer to have `py39-trio` rather than `notrio` 😁

njsmith:
Fair enough!

njsmith:
Made that change and also rebased on latest master, to make mergify happy

.circleci/config.yml
```diff
@@ -64,6 +64,15 @@ jobs:
...
             tox -e py39
+  py39-notrio:
+    docker:
+      - image: circleci/python:3.9
+    steps:
+      - checkout
+      - run:
+          command: |
+            sudo pip install tox
+            tox -e py39-notrio
```
penguinolog, line 75: inconsistent with `tox.int` (`py39-trio` for `trio` and normal python versions for other)

chrisjwelly:
Hello @njsmith ! I understand it has been over 2 years since this PR was opened. Might there be any chance of a follow-up to support trio out of the box? 

I think this feature is great, as it can be easy to forget add `trio.sleep` argument to the retry.

SysuJayce:
I also wonder how to use tenacity with trio.

If I use tenacity.retry() as a decorator for a async function, how can I check the result of the async function via `tenacity.retry_if_result()`? And then I got an error `TypeError: argument of type 'coroutine' is not iterable, xxx was never awaited`.

Even if I change my retry condition to `retry_if_exception`, I got another error `TypeError: object NoneType can't be used in 'await' expression`.

Please give me some tips, thank you guys~~~

jakkdl:
This can be closed, as it was implemented with #463

Issue #137, opened 2018-08-10
Retry being called despite successful execution of function


chetnaB:
On using `retry_unless_exception_type` , retry is always called (despite a successful run) until the exception(s) is(are) encountered. However, correct behavior with this condition should be to retry **only if a function has failed due to some exception other than the ones mentioned** in the retry condition.
e.g. This is the function definition, wherein I want to retry the function only if the exception is **not** among `psycopg2.ProgrammingError` and `NameError` and either the time since last retrial is less than 1 min, or the attempt count is less than 3:
```
@retry(stop=(stop_after_delay(60) | stop_after_attempt(2)),
        reraise=True,
        retry=retry_unless_exception_type((psycopg2.ProgrammingError, NameError)))
def run_query(query_variable):
    cursor.execute(query_variable)
    print cursor.rowcount
    print "Done"
```

and here's the execution log, which shows that function was retried twice despite successful runs:
```
53059
Done
53059
Done
Traceback (most recent call last):
  File "update_offer_base_item_mapping.py", line 22, in <module>
    run_query(a)
  File "/Users/grofers/Documents/cb/py2venv/lib/python2.7/site-packages/tenacity/__init__.py", line 241, in wrapped_f
    return self.call(f, *args, **kw)
  File "/Users/grofers/Documents/cb/py2venv/lib/python2.7/site-packages/tenacity/__init__.py", line 330, in call
    start_time=start_time)
  File "/Users/grofers/Documents/cb/py2venv/lib/python2.7/site-packages/tenacity/__init__.py", line 297, in iter
    raise retry_exc.reraise()
  File "/Users/grofers/Documents/cb/py2venv/lib/python2.7/site-packages/tenacity/__init__.py", line 137, in reraise
    raise self
tenacity.RetryError: RetryError[<Future at 0x109cf4190 state=finished returned NoneType>]
```

jd:
You want `retry=retry_unless_exception_type((psycopg2.ProgrammingError, NameError)) | retry_if_result(lambda r: True))` AFAIU.

yuvadm:
@jd this is a serious pitfall, I had the same expectation about `retry_unless_exception_type` and got bit by this awkward behavior.

asqui:
Hmmm, so it seems the problem here is that there is an ambiguity in the interpretation of:
`retry=retry_unless_exception_type(T1, T2)`

When reading this code, there are two possible interpretations:
1. Retry until an exception of type `T1` or `T2` is raised.
[...]

PR #300, opened 2021-05-17
Add retry_if_not_exception_type()

### LINKED ISSUE #256: Retry if not exception type?
Hi, I was wondering if there's an easy way to say retry for everything except this list of exceptions?

For example, I want to retry for all server errors but not client errors.

henry0312:
Fixes #256

### DIFF
--- a/releasenotes/notes/add-retry_except_exception_type-31b31da1924d55f4.yaml
+++ b/releasenotes/notes/add-retry_except_exception_type-31b31da1924d55f4.yaml
@@ -1,0 +1,3 @@
+---
+features:
+  - Add ``retry_except_exception_type()`` that allows us to retry if a raised exception doesn't match given exceptions.
--- a/releasenotes/notes/add-retry_except_exception_type-31b31da1924d55f4.yaml
+++ b/releasenotes/notes/add-retry_except_exception_type-31b31da1924d55f4.yaml
@@ -1,3 +1,3 @@
 ---
 features:
-  - Add ``retry_except_exception_type()`` that allows us to retry if a raised exception doesn't match given exceptions.
+  - Add ``retry_except_exception_type()`` that allows to retry if a raised exception doesn't match given exceptions.
--- a/releasenotes/notes/add-retry_except_exception_type-31b31da1924d55f4.yaml
+++ b/releasenotes/notes/add-retry_except_exception_type-31b31da1924d55f4.yaml
@@ -1,3 +1,3 @@
 ---
 features:
-  - Add ``retry_except_exception_type()`` that allows to retry if a raised exception doesn't match given exceptions.
+  - Add ``retry_if_not_exception_type()`` that allows to retry if a raised exception doesn't match given exceptions.
--- a/tenacity/retry.py
+++ b/tenacity/retry.py
@@ -80,6 +80,16 @@ class retry_if_exception_type(retry_if_exception):
         )
 
 
+class retry_except_exception_type(retry_if_exception):
+    """Retries excpet an exception has been raised of one or more types."""
+
+    def __init__(self, exception_types=Exception):
+        self.exception_types = exception_types
+        super(retry_except_exception_type, self).__init__(
+            lambda e: not isinstance(e, exception_types)
+        )
+
+
 class retry_unless_exception_type(retry_if_exception):
     """Retries until an exception is raised of one or more types."""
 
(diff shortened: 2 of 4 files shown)


releasenotes/notes/add-retry_except_exception_type-31b31da1924d55f4.yaml
```diff
@@ -0,0 +1,3 @@
+---
+features:
+  - Add ``retry_except_exception_type()`` that allows us to retry if a raised exception doesn't match given exceptions.
```
jd, line 3:
```suggestion
  - Add ``retry_except_exception_type()`` that allows to retry if a raised exception doesn't match given exceptions.
```

tenacity/retry.py
```diff
@@ -80,6 +80,16 @@ def __init__(self, exception_types=Exception):
         )
 
 
+class retry_except_exception_type(retry_if_exception):
+    """Retries excpet an exception has been raised of one or more types."""
```
jd, line 84:
```suggestion
    """Retries except an exception has been raised of one or more types."""
```

jd (changes requested):
Small typo

jd (dismissed):

asqui:
I think we can be clearer on the naming of this method to better communicate what it is doing.

Is this implementing the `retry_if_exception_not_of_type` which was propossed in the discussion of #137?

henry0312:
> Is this implementing the retry_if_exception_not_of_type which was propossed in the discussion of #137?

Yes, it it.
Though @jd approved this, would it be good to rename `retry_except_exception_type` to `retry_if_exception_not_of_type`?

jd:
I'm fine with whatever.
We need to fix the CI not running issue to merge anything at this point 😢

asqui:
I do think `retry_if_exception_not_of_type` would fit better amongst the existing family of related `retry_*` methods:

- `retry_if_exception`
- `retry_if_exception_type`
- `retry_unless_exception_type` -- would have been better named `retry_if_not_exception_type`, but yolo.
- `retry_if_result`
- `retry_if_not_result`
- `retry_if_exception_message`
- `retry_if_not_exception_message`

In particular, I think it's important that the new method, whatever it's called, is clearly set apart from `retry_unless_exception_type` given the subtle but important differences between the two (as discussed, at length, in #137).

More generally, I'm slightly alarmed by the "I'm fine with whatever" remark. I worry that without careful consideration of usability, adoption of this library will be hampered in the future. (This is speaking as someone who has tried to bring tenacity in to an existing codebase which already has numerous disparate approaches to retry logic, and the complexity/usability concerns around tenacity have certainly been a barrier to adoption.)

henry0312:
A relation between `retry_if_exception_type()` and `retry_except_exception_type()` is the same with one between `retry_if_exception_message()` and `retry_if_not_exception_message()`, since the implementation of `retry_except_exception_type()` is the essentially same with that of `retry_if_not_exception_message()`.
(cf. https://github.com/jd/tenacity/blob/master/tenacity/retry.py#L125 and https://github.com/jd/tenacity/blob/master/tenacity/retry.py#L167)

Therefore, if I would rename, I think `retry_if_not_exception_type` is better than `retry_if_exception_not_of_type`, that follows the named rule of `retry_if_not_exception_message`.

---
For details,

- `retry_if_exception_type(T)`

`T` is **included** in conditons of retry.

- `retry_except_exception_type(T)`

`T` is **excluded** from conditions of retry.
In other words, `T` is **not included** in condtions of retry.

- `retry_if_exception_message(A)`

`A` is **included** in conditons of retry.

- `retry_if_not_exception_message(A)`

`A` is **excluded** from conditions of retry.
This means that `T` is **not included** in condtions of retry.

By the way,

- `retry_unless_exception_type(T)`

Retry **until** `T` is raised.
(`T` is expected to be raised to stop retry)

---

I'm not good ad English and I'm sorry for my bad English.

asqui:
@henry0312 Thanks for taking the time to look at this in such detail.

I agree that `retry_if_not_exception_type` fits better with the existing naming conventions.

P.S. Don't apologise, your english is great! (I have no reason to believe you're not a native speaker!)

henry0312:
@jd @asqui I just renamed `retry_except_exception_type` to `retry_if_not_exception_type`

jd (approved):

henry0312:
@jd I just fixed tests!

jd (approved):

henry0312:
@jd Thank you for reviewing and merging.
Do you have a plan for a new [release](https://github.com/jd/tenacity/releases)?

william-silversmith:
Likewise! Thanks so much jd and all that contributed! I would start using this feature once the next release is out.

PR #302, opened 2021-06-06
Add a __repr__ method to RetryCallState objects

zbentley:
Many folks use tools which dump locals when an exception occurs. For example, I use Sentry quite frequently to dig around in the stack below an exception to see what happened and what went wrong. Many of those stack-debugging tools stringify (via `__repr__`ing each local) and send the results to some sort of debugging console/webpage.

When I encounter a Tenacity-retried function in my call stack, I can poke around inside Tenacity's locals a bit to see what went wrong, but I can't tell much about what caused Tenacity to give up, since all I can see of the retry call state object is e.g. `<tenacity.RetryCallState object at 0x7f99fe9119a0>`. That's not terribly useful.

This PR adds a `__repr__` method to RetryCallState so that stringifications of those objects now contain some information useful for human debuggers--how many times something has been retried, how long Tenacity has slept for, and what the result of the last attempt (if any) was.


jd (changes requested):
That's a good idea. Would you mind adding a changelog entry and some tests?

zbentley:
@jd I am happy to; mobile until the 19th and then will pick this up. Thanks for the response and excellent library!

jd (approved):

jd (approved):

zbentley:
@jd something about my tests appears to be merge conflicting. Did I perhaps run the wrong version of `black` on them or something similar?

penguinolog:
`black` changed formatting in 2021 year versions

penguinolog (commented):
@zbentley `tenacity/tests/test_tenacity.py` has been moved to `tests/test_tenacity.py`, it's reason for merge conflicts

zbentley:
@penguinolog all set, thanks!

penguinolog (approved):

jd (approved):

PR #308, opened 2021-06-23
Do not package tests with tenacity

penguinolog:
* move tests outside of package
* fix DeprecationWaring for `unittest` API (in tests)
* Do not raise exceptions classes, but raise exception instances


.editorconfig
```diff
@@ -12,7 +12,7 @@ trim_trailing_whitespace = true
 indent_size = 4
 max_line_length = 120
 
-[*.ini]
+[*.{ini,cfg}]
```
penguinolog, line 15: figured, that current `setup.cfg` is indented with 4 spaces -> do not reformat

tenacity/_asyncio.py
```diff
@@ -58,7 +58,7 @@ async def __anext__(self):
         while True:
             do = self.iter(retry_state=self._retry_state)
             if do is None:
-                raise StopAsyncIteration
+                raise StopAsyncIteration()
```
jd, line 61: unrelated?

jd (changes requested):

tenacity/_asyncio.py
penguinolog: Formally bug, but yes: unrelated. It should be separate change

and-semakin (approved):
Looks good!

jd (dismissed):
There are a lot of unrelated changes. It'd really be great to split this in multiple PRs.

penguinolog:
Removed all changes except move & fix imports

jd (approved):

PR #309, opened 2021-06-24
Remove encoding declarations

and-semakin:
They are useless for Python 3 code. Everything is UTF-8 by default.


and-semakin:
I don't add a changelog entry because it could be treated as a part of Python 2 deprecation.

penguinolog (approved):

jd (approved):

PR #310, opened 2021-06-24
Replace abc.ABCMeta with abc.ABC

and-semakin:
`abc.ABC` is available in Python 3.4+ and is a simpler alternative to `abc.ABCMeta`. As `tenacity` no longer supports Python 2.7, we could use it.


and-semakin:
I don't add a changelog entry because it could be treated as a part of Python 2 deprecation.

jd (approved):

PR #311, opened 2021-06-24
Use f-strings instead of `str.format` as faster and easier to read

penguinolog:
Since python 3.6 we are able to use it.
Use attributes in `repr` directly instead of chained construction
 -> remove unused private helper

no-changelog

### DIFF
--- a/tenacity/before_sleep.py
+++ b/tenacity/before_sleep.py
@@ -38,7 +38,7 @@ def before_sleep_log(
     def log_it(retry_state: "RetryCallState") -> None:
         if retry_state.outcome.failed:
             ex = retry_state.outcome.exception()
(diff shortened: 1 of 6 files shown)


tenacity/before_sleep.py
```diff
@@ -50,12 +50,8 @@ def log_it(retry_state: "RetryCallState") -> None:
 
         logger.log(
             log_level,
-            "Retrying {0} in {1} seconds as it {2} {3}.".format(
-                _utils.get_callback_name(retry_state.fn),
-                getattr(retry_state.next_action, "sleep"),
```
penguinolog, line 55: `getattr` here not protect from `AttributeError` exceptions since no default value specified -> use attribute directly

jd (approved):

Issue #288, opened 2021-04-07
Bug: decorators on async funcs change their __name__


WestXu:
tenacity version 7.0.0

```python
async def test(self):
    pass

@tenacity.retry
def test2(self):
    pass

@tenacity.retry
async def test3(self):
    pass

print(test.__name__, test2.__name__, test3.__name__)
```
produces `test test2 async_wrapped`

```python
class T:
    async def test(self):
        pass
    
    @tenacity.retry
    def test2(self):
        pass
    
    @tenacity.retry
    async def test3(self):
        pass
    
t = T()
print(t.test.__name__, t.test2.__name__, t.test3.__name__)
```
produces `test test2 async_wrapped`

`test test2 test3` is expected in both cases.

WestXu:
Checked, works fine at 6.2.0, broke at 6.3.0.

penguinolog:
fixed

jd:
@penguinolog if you know the PR don't forget to say how it was fixed 😉

PR #312, opened 2021-06-24
Fix issue #288 : __name__ and other attributes for async functions

penguinolog:
`iscoroutinefunction` is always not Nons -> remove check

### DIFF
--- a/tenacity/__init__.py
+++ b/tenacity/__init__.py
@@ -118,7 +118,7 @@ def retry(*dargs: t.Any, **dkw: t.Any) -> t.Union[WrappedFn, t.Callable[[Wrapped
                     f"Got retry_base instance ({f.__class__.__name__}) as callable argument, "
                     f"this will probably hang indefinitely (did you mean retry={f.__class__.__name__}(...)?)"
                 )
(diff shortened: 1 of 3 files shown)


tenacity/__init__.py
```diff
@@ -118,7 +118,7 @@ def wrap(f: WrappedFn) -> WrappedFn:
                     f"Got retry_base instance ({f.__class__.__name__}) as callable argument, "
                     f"this will probably hang indefinitely (did you mean retry={f.__class__.__name__}(...)?)"
                 )
-            if iscoroutinefunction is not None and iscoroutinefunction(f):
```
penguinolog, line 121: 1 part is always `True`

tenacity/_asyncio.py
```diff
@@ -70,6 +72,7 @@ def wraps(self, fn):
         fn = super().wraps(fn)
         # Ensure wrapper is recognized as a coroutine function.
 
+        @functools.wraps(fn)
```
penguinolog, line 75: bugfix...

jd (approved):

PR #315, opened 2021-06-24
Add type annotations to cover all code.

penguinolog:
`tenacity` is marked as "typed" package,
which means all public APIs should be annotated


jd:
Could we also run mypy on the ci?
That'd make sure what we provide here is correct.

jd (approved):

PR #317, opened 2021-06-30
Fix after_log logger format

penguinolog:
Restored PR #299 from @akgnah with fixed conflicts and addressed comments

PR #319, opened 2021-07-05
Fix after_log message

RoyalTS:
This seems to have been mangled a bit in the recent move to f-strings


jd (approved):
Worth a release note entry?

RoyalTS:
This strikes me as super minor and I don't want to faff around with reno unless I absolutely have to. Feel free to merge as is.

jd:
Fixed with #317

PR #325, opened 2021-07-16
Add dynamic configuration for stop & wait conditions

gtors:
This PR adds the ability to configure `@retry` based on the attributes of a class instance

Simplified use case:
 ```py
class Service:
    def __init__(self, tries):
        self.tries = tries

    @retry(
        retry=retry_if_exception_type(IOError),
        stop=stop_after_attempts(attempts=from_attr("tries")),
    )
    def call_other_service(self):
        ...


def load_config_from_file_system():
    ...


def main():
    config = load_config_from_file_system()
    serivice = Service(config.service.retry_attempts)
    serivice.call_other_service()
```

### DIFF
--- a/releasenotes/notes/dynamic-configuration-75ab1a482dd2e5b7.yaml
+++ b/releasenotes/notes/dynamic-configuration-75ab1a482dd2e5b7.yaml
@@ -1,0 +1,3 @@
+---
+features:
+  - Add the ability to dynamically configure the stop and wait conditions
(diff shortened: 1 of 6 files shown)


penguinolog:
looks like it's job for descriptor, not normal decorator. In case of descriptor, owner class will be registered during class construction and owner instance will be sent to `Retrying` instance. But this is serious change

jd:
Closing this as it's quite old and conflicting.

PR #339, opened 2022-01-23
add mypy check on CI

hirosassa:
Related to https://github.com/jd/tenacity/pull/315#issuecomment-871366173

In this PR, I added mypy check on CI and fixed easy type error (ignored for complicated errors, will be fixed in another PR).

### DIFF
--- a/tenacity/after.py
+++ b/tenacity/after.py
@@ -36,6 +36,8 @@ def after_log(
     """After call strategy that logs to some logger the finished attempt."""
 
     def log_it(retry_state: "RetryCallState") -> None:
+        if retry_state.fn is None:
+            return None
         logger.log(
             log_level,
             f"Finished call to '{_utils.get_callback_name(retry_state.fn)}' "
--- a/tenacity/before_sleep.py
+++ b/tenacity/before_sleep.py
@@ -36,17 +36,26 @@ def before_sleep_log(
     """Before call strategy that logs to some logger the attempt."""
 
     def log_it(retry_state: "RetryCallState") -> None:
+        if retry_state.outcome is None:
+            return None
+
         if retry_state.outcome.failed:
             ex = retry_state.outcome.exception()
             verb, value = "raised", f"{ex.__class__.__name__}: {ex}"
 
             if exc_info:
-                local_exc_info = retry_state.outcome.exception()
+                local_exc_info = retry_state.outcome.exception() is not None
             else:
                 local_exc_info = False
         else:
             verb, value = "returned", retry_state.outcome.result()
             local_exc_info = False  # exc_info does not apply when no exception
+
+        if retry_state.next_action is None:
+            return None
+
+        if retry_state.fn is None:
+            return None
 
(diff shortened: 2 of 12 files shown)


jd (dismissed):

jd:
LGTM but circleci is broken again :(

hirosassa:
Thanks for the review! I added mergify config.

> circleci is broken

Should I just wait for https://github.com/jd/tenacity/pull/340 to be merged? or Can I fix this?

ysk24ok:
@jd Hello. Excuse me for cutting in.
What's the status of this PR? There's something I'd like to work on after this PR is merged.

hirosassa:
@ysk24ok Thanks for the comment!
@jd Hi, jd. Could we merge this PR?

jd (approved):

jd:
Not sure why tests fail?

tenacity/before_sleep.py
```diff
@@ -36,18 +36,27 @@ def before_sleep_log(
...
+        if retry_state.outcome is None:
+            return None
+
         if retry_state.outcome.failed:
             ex = retry_state.outcome.exception()
             verb, value = "raised", f"{ex.__class__.__name__}: {ex}"
 
             if exc_info:
-                local_exc_info = retry_state.outcome.exception()
+                local_exc_info = retry_state.outcome.exception() is not None
```
ysk24ok, line 47: This change breaks the existing test (`TestBeforeAfterAttempts.test_before_sleep_log_raises_with_exc_info`).
The `local_exc_info` is passed to `logger.log` so it should not be bool. 
Instead, the value should be the exception itself or None.
https://docs.python.org/3/library/logging.html#logging.LogRecord

tenacity/after.py
```diff
@@ -36,6 +36,8 @@ def after_log(
     """After call strategy that logs to some logger the finished attempt."""
 
     def log_it(retry_state: "RetryCallState") -> None:
+        if retry_state.fn is None:
+            return None
```
ysk24ok, line 40: This change breaks the existing tests (`TestAfterLogFormat.test_01_default` and `TestAfterLogFormat.test_02_custom_sec_format`).
In the real use case, `retry_state.fn` is not None (it should be the decorated function), but in tests the object is created by `make_retry_state` and the `fn` attribute is None by default.
Maybe we should manually add a dummy function to `fn` or we may use `Retrying` object to mock the decorator behavior.

tenacity/before_sleep.py
hirosassa: @ysk24ok Thanks for your comments! I'll take a look tonight.

dgutson:
Hi, any update on this? @hirosassa do you need help?

hirosassa:
@dgutson Thanks for reminding me! I'll fix this PR this weekend.

sileht:
Thanks, @hirosassa for this, this got incorporated in https://github.com/jd/tenacity/pull/382.

PR #340, opened 2022-01-24
ci: fix Mergify config

jd:


sileht:
@mergifyio refresh

sileht:
@Mergifyio refresh

PR #342, opened 2022-01-24
Support `datetime.timedelta` as a valid wait unit type

### LINKED ISSUE #341: Support timedelta as a valid type for `wait` classes
Currently, the `wait`ing classes only take `int` or `float` as their wait unit types where they are representing seconds.
It would make for a better, more readable code if I could simply pass in a `datetime.timedelta` instead and thus also avoid passing in `wait=60*5` for 5 minutes and instead could `wait=timedelta(minutes=5)`

noambloom:
Closes #341

### DIFF
--- a/tenacity/wait.py
+++ b/tenacity/wait.py
@@ -17,11 +17,18 @@
 import abc
 import random
 import typing
+from datetime import timedelta
 
 from tenacity import _utils
 
 if typing.TYPE_CHECKING:
     from tenacity import RetryCallState
+
+wait_unit_type = typing.Union[int, float, timedelta]
+
+
+def to_seconds(wait_unit: wait_unit_type) -> float:
+    return float(wait_unit.total_seconds() if isinstance(wait_unit, timedelta) else wait_unit)
 
 
 class wait_base(abc.ABC):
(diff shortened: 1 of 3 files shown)


noambloom:
@jd do you mind taking a look please? Thanks :)

doc/source/index.rst
```diff
@@ -213,9 +213,9 @@ exceptions, as in the cases here.
         print("Retry forever with no wait if an IOError occurs, raise any other errors")
         raise Exception
 
-    @retry(retry=retry_if_not_exception_type(ClientError))
+    @retry(retry=retry_if_not_exception_type(ConnectionError))
     def might_client_error():
-        print("Retry forever with no wait if any error other than ClientError occurs. Immediately raise ClientError.")
+        print("Retry forever with no wait if any error other than ConnectionError occurs. Immediately raise ConnectionError.")
```
jd, line 218: This seems unrelated and has been fixed in a different PR.

jd (dismissed):

tenacity/wait.py
```diff
@@ -141,14 +148,14 @@ class wait_exponential(wait_base):
 
     def __init__(
         self,
-        multiplier: typing.Union[int, float] = 1,
-        max: typing.Union[int, float] = _utils.MAX_WAIT,  # noqa
+        multiplier: wait_unit_type = 1,
```
jd, line 151: This is not time, just a multiplier.

jd (dismissed):
Might be nice to have a couple of tests also, even if typing makes it work in theory.

jd (dismissed):

jd (approved):

PR #347, opened 2022-03-23
Explicitly export convenience symbols from tenacity/__init__.py

tomasgareau:
Hello! This PR tweaks all of the convenience exports in `tenacity/__init__.py` to use the `import ... as ...` form. This lets type-checkers discover those symbols under the `tenacity` module, e.g.:

**Before**
```python
import tenacity

tenacity.wait_random_exponential()  # Pyright: "wait_random_exponential" is not a known member of module
```

**After**
```python
import tenacity

tenacity.wait_random_exponential()  # Pyright: (class) wait_random_exponential
```

Pyright (and I _think_ mypy) will consider [imported symbols as private by default](https://github.com/microsoft/pyright/blob/main/docs/typed-libraries.md#library-interface) unless they:
* use the `import X as X` or `from A import X as X` form, (a redundant alias) or,
* include the name of the symbol in an `__all__` symbol

The `from A import X as X` form was easier for me to do with a macro but happy to change it to the `__all__` form if you'd prefer. Just thought I'd kick off this PR as a place to discuss that.


asqui:
Personally I think defining `__all__` makes it clearer that the intent is to define a public API.

I'm not familiar with the import conventions you mentioned or their origins, but to me they wouldn't make the intent obvious, and the 'redundant alias' seems like a subtle code smell that's liable to be "cleaned up" by some well-meaning individual in the future.

tomasgareau:
I think the import convention of "import ... as ..." comes (loosely) from [PEP-484's section on stub files](https://peps.python.org/pep-0484/#stub-files):

> Modules and variables imported into the stub are not considered exported from the stub unless the import uses the import ... as ... form or the equivalent from ... import ... as ... form. (UPDATE: To clarify, the intention here is that only names imported using the form X as X will be exported, i.e. the name before and after as must be the same.)

I say loosely because the typing in here is in-line, not in a stub file. However, it seems as though type-checkers (or at least Pyright) [use the same convention for in-line typing too](https://github.com/microsoft/pyright/blob/main/docs/typed-libraries.md#library-interface) (which makes sense -- more consistent that way). I've found that Pyright's automatic stub generator will use the `from X import A as A` form to export symbols.

That said, I agree with you that the `__all__` form is more explicit -- just force pushed dd86b81 to switch to using `__all__`.

tenacity/__init__.py
```diff
@@ -513,3 +513,58 @@ def __repr__(self):
...
+    "DoSleep",
+    "BaseAction",
+    "RetryAction",
+    "RetryError",
+    "AttemptManager",
+    "BaseRetrying",
+    "Retrying",
+    "Future",
+    "RetryCallState",
+    "AsyncRetrying",
```
tomasgareau, line 569: In addition to the convenience exports in lines 519-555, I also added these classes which were publicly defined in `__init__.py`. Not super familiar with `tenacity` though so someone might want to check that these are intended for `tenacity`'s public API (or if there are any symbols missing here).

bhrutledge:
If this could be updated to pass CI, think it would fix #230.

tenacity/__init__.py
sileht: They are.

sileht (approved):

PR #355, opened 2022-06-02
Allow asynchronous callbacks for async retries.

### LINKED ISSUE #249: Is an async retry_error_callback possible?

```python

    print(retry_state)

@retry(
        stop=stop_after_attempt(1),
        retry_error_callback=relogin
    )
        return response["execId"]
```

Sadly, I got

```
```

mastizada:
This change will allow both async and sync functions to be used as callable parameters for Async retries.

Docs are not updated, please, inform if any other changes are needed. Thank you very much for your review.

Fixes #249

### DIFF
--- a/tenacity/_asyncio.py
+++ b/tenacity/_asyncio.py
@@ -17,35 +17,31 @@
 
 import functools
 import sys
-import typing
-from asyncio import sleep
+from asyncio import sleep as aio_sleep
+from inspect import iscoroutinefunction
+from typing import Any, Awaitable, Callable, TypeVar, Union
 
-from tenacity import AttemptManager
-from tenacity import BaseRetrying
-from tenacity import DoAttempt
-from tenacity import DoSleep
-from tenacity import RetryCallState
+from tenacity import AttemptManager, BaseRetrying, DoAttempt, DoSleep, RetryAction, RetryCallState, TryAgain
 
-WrappedFn = typing.TypeVar("WrappedFn", bound=typing.Callable)
-_RetValT = typing.TypeVar("_RetValT")
+WrappedFn = TypeVar("WrappedFn", bound=Callable)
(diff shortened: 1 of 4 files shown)


tenacity/_asyncio.py
```diff
@@ -17,35 +17,31 @@
...
 
-WrappedFn = typing.TypeVar("WrappedFn", bound=typing.Callable)
-_RetValT = typing.TypeVar("_RetValT")
+WrappedFn = TypeVar("WrappedFn", bound=Callable)
+_RetValT = TypeVar("_RetValT")
 
 
 class AsyncRetrying(BaseRetrying):
-    def __init__(self, sleep: typing.Callable[[float], typing.Awaitable] = sleep, **kwargs: typing.Any) -> None:
+    def __init__(self, sleep: Callable[[float], Awaitable] = aio_sleep, **kwargs: Any) -> None:
```
jd, line 31: There are too many changes that are unrelated to this PR. 
Could you avoid changing code that does not need to be changed?

jd (changes requested):

tenacity/_asyncio.py
mastizada: It was giving warnings because variable and function had same names (`sleep`), should I make another PR for this?

mastizada: Also, writing `typing.` before types every time made some lines too long, and it was looking unnecessary, should I revert that one too?

mastizada: sorry for late response, I will try to find time in upcoming days and split this PR into 2 different PRs.

mastizada: I created a new PR with AsyncRetries related changes #363 
Will make another PR for coding style improvements after that.

Closing this PR (will use this branch as template)

Issue #356, opened 2022-07-03
PEP-541: Publish under retrying on PyPI


groodt:
Would the maintainers of this package be interested in obtaining the abandoned `retrying` name on PyPI? This is possible by either asking the current owner or through PEP-541.

I see the following benefits:
* Existing users of retrying can use a maintained package with fixes and updates and improvements 

I don't believe we need to be concerned with breaking changes:
* A compatibility shim can be included for the imports
* A new version would be published with a breaking version number and if anyone was not pinning they could be free to do so

groodt:
I suggest claiming the name on `pypi` via https://github.com/pypa/pypi-support/issues/new?labels=PEP+541&template=pep541-request.yml&title=PEP+541+Request%3A+PROJECT_NAME

groodt:
There is a similar 541  request happening here. We can keep an eye on that approach and see how it plays out. https://github.com/pypa/pypi-support/issues/2064

bmrobin:
@groodt @jd https://github.com/pypi/support/issues/2064 appears resolved and ownership transfer was completed. do either of you have any interest in continuing with this proposal for `tenacity`?

jd:
Tenacity deliberately chose a different name and is well-established under it. Taking over the `retrying` name on PyPI would create more confusion than it solves. Closing as not planned.

PR #359, opened 2022-07-20
Add retrying shim

groodt:
Introduces a shim that should enable drop-in replacement for the unmaintained https://github.com/rholder/retrying project.

Path towards fix for:
https://github.com/jd/tenacity/issues/356

An alternative approach might be to claim the name on PyPI and then publish a dummy package (with a breaking version number) with the shim. Over time you could standardise on one name or the other and log warnings and eventually fail (with another breaking version number).

### DIFF
--- a/retrying/__init__.py
+++ b/retrying/__init__.py
@@ -1,2 +1,1 @@
-# flake8: noqa
-from tenacity import retry
+from tenacity import retry # noqa
--- a/retrying/__init__.py
+++ b/retrying/__init__.py
@@ -1,0 +1,2 @@
+# flake8: noqa
+from tenacity import retry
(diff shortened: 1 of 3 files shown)


jd:
Not sure this fixes/help with anything though?

groodt:
Yes, agree. Was thinking that it might assist individuals migrating to tenacity without changing imports, but it probably isn’t necessarily the best strategy. 

Taking a step back, what would you like to see happen with “retrying” on pypi. It’s presently abandoned so is a supply chain risk / unmaintained code. 

If you want the pypi project name, you should submit a 541 request because they take 2 or 3 months to process. https://github.com/pypa/pypi-support/issues/2064#issuecomment-1194395022

Some ideas for how to migrate pypi packages.

Option 1: Over time, migrate tenacity users to retrying

Option 2: Over time, migrate retrying users to tenacity

Option 3: Keep users of both packages as-is, but without maintaining 2 packages and code based long-term (I’m not sure this is possible or desirable)

PR #362, opened 2022-09-20
feat: Add retry_if_exception_cause_type

Greesb:
Add a new retry_base class called `retry_if_exception_cause_type` that checks that the cause of the raised exception is of a certain type.

### DIFF
--- a/tenacity/retry.py
+++ b/tenacity/retry.py
@@ -117,6 +117,33 @@ class retry_unless_exception_type(retry_if_exception):
         return self.predicate(retry_state.outcome.exception())
 
 
+class retry_if_exception_cause_type(retry_base):
+    """Retries if any of the causes of the raised exception is of one or more types.
+
+    The check on the type of the cause of the exception is done recursively (until finding
+    an exception in the chain that has no `__cause__`)
+    """
+
+    def __init__(
+        self,
+        exception_types: typing.Union[
+            typing.Type[BaseException],
+            typing.Tuple[typing.Type[BaseException], ...],
+        ] = Exception,
+    ) -> None:
+        self.exception_cause_types = exception_types
+
+    def __call__(self, retry_state: "RetryCallState") -> bool:
+        if retry_state.outcome.failed:
+            exc = retry_state.outcome.exception()
+            while exc is not None:
+                if isinstance(exc.__cause__, self.exception_cause_types):
+                    return True
+                exc = exc.__cause__
+
+        return False
+
+
 class retry_if_result(retry_base):
     """Retries if the result verifies a predicate."""
 
--- a/releasenotes/notes/add_retry_if_exception_cause_type-d16b918ace4ae0ad.yaml
+++ b/releasenotes/notes/add_retry_if_exception_cause_type-d16b918ace4ae0ad.yaml
@@ -1,0 +1,5 @@
+---
+features:
+  - |
+    Add a new `retry_base` class called `retry_if_exception_cause_type` that
+    checks, recursively, if any of the causes of the raised exception is of a certain type.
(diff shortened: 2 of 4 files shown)


sileht (dismissed):

JulesVautier (commented):
Do we add a releasenote if the change doesn't impact customers ?

JulesVautier (approved):

tenacity/retry.py
```diff
@@ -117,6 +117,29 @@ def __call__(self, retry_state: "RetryCallState") -> bool:
         return self.predicate(retry_state.outcome.exception())
 
 
+class retry_if_exception_cause_type(retry_base):
+    """Retries if the cause of the raised exception is of one or mor types."""
```
asqui, line 121: Typo in “…one or mor types”

asqui (commented):
Is this change really necessary?

Why not just use `retry_if_exception(predicate)` with a lambda to achieve the same effect?

Greesb:
> Is this change really necessary?
> 
> Why not just use `retry_if_exception(predicate)` with a lambda to achieve the same effect?

By "this change" do you mean the way it is implemented ? Or do you mean the new class on its own ?

If you're talking about the new class on its own, then a lot of existing `retry_base` class can be achieved with a predicate with lambda, it's just a quality of life improvement to already have a "native" feature that does it to not have to implement a lambda that does it.

If you're talking about the way it is implemented, then yeah it could also probably be done with a predicate, but it'll probably look really ugly. I just thought it looked better like this.

asqui:
Sorry for not being clear, and thank you for your contribution.

I meant "why do we need this class", because I didn't see a big difference between:
```
retry_if_exception(lambda e: isinstance(e.cause, SomeException))
```
and
```
retry_if_exception_cause_type(SomeExcetion)
```
However, on closer inspection, I see that this is slightly different because your implementation considers all chained exceptions and not just the first.

I guess this is quite deliberate so it would be good to make this more clear in the docstring/release note -- my initial feedback was based on the docstring/release note from which I assumed it to be equivalent to the first snippet above.

(I don't think that "exception cause" makes it immediately obvious that this includes all causes up the chain of nested exceptions; at least not to me.)

As for the implementation, I don't think using `retry_if_exception` and a predicate will add or take away much (would remove the if statement and the `return False` but add a `def inner():` and `super()` in its place, no big deal either way) though I do see that all of the other `retry_*exception_*` classes follow the pattern of inheriting from `retry_if_exception` so this would be the first deviation from that...

Greesb:
> Sorry for not being clear, and thank you for your contribution.
> 
> I meant "why do we need this class", because I didn't see a big difference between:
> 
> ```
> retry_if_exception(lambda e: isinstance(e.cause, SomeException))
> ```
> 
> and
> 
> ```
> retry_if_exception_cause_type(SomeExcetion)
> ```
> 
> However, on closer inspection, I see that this is slightly different because your implementation considers all chained exceptions and not just the first.
> 
> I guess this is quite deliberate so it would be good to make this more clear in the docstring/release note -- my initial feedback was based on the docstring/release note from which I assumed it to be equivalent to the first snippet above.
> 
> (I don't think that "exception cause" makes it immediately obvious that this includes all causes up the chain of nested exceptions; at least not to me.)
> 
> As for the implementation, I don't think using `retry_if_exception` and a predicate will add or take away much (would remove the if statement and the `return False` but add a `def inner():` and `super()` in its place, no big deal either way) though I do see that all of the other `retry_*exception_*` classes follow the pattern of inheriting from `retry_if_exception` so this would be the first deviation from that...

You're right the documentation was a bit unclear on what it was doing, i updated the docstring and the release note.
Let me know if it's better and/or if you've got any other suggestions.

asqui:
Nice, that’s a lot clearer. No further feedback from me.

AIGeneratedUsername:
Could you please update https://tenacity.readthedocs.io/en/latest/changelog.html alongside with new releases? The "8.1.0" is the latest at Pypi, but I see no changelog changes. Thanks.

PR #363, opened 2022-09-23
Allow asynchronous callbacks for AsyncRetrying parameters

mastizada:


jd (commented):
Could we try to DRY the `iter` method? I understand we need to re-define it but it'd be great if we could share most of the code.

tenacity/_asyncio.py
```diff
@@ -90,3 +93,48 @@ async def async_wrapped(*args: typing.Any, **kwargs: typing.Any) -> typing.Any:
...
+
+    @staticmethod
+    async def handle_custom_function(
+        func: typing.Union[typing.Callable, typing.Awaitable], retry_state: RetryCallState
+    ) -> typing.Any:
+        if iscoroutinefunction(func):
+            return await func(retry_state)
+        return func(retry_state)
+
+    async def iter(self, retry_state: "RetryCallState") -> typing.Union[DoAttempt, DoSleep, typing.Any]:  # noqa
```
mastizada, line 105: @jd I also wanted to use some base method to avoid re-defining the iter method again, however, one of them is sync `def iter` and other one is async `async def iter`. Any ideas how to make it DRY?

I wanted to use `handle_custom_function` to handle sync/async calls, however, if we make `handle_custom_function` sync function, then it might need to use `asyncio.get_event_loop()` to run awaitable functions (running async functions from sync function).

mastizada: @jd any ideas? I would love to have progress with this PR and start using async callbacks for `retry_error_callback`

PR #366, opened 2022-10-09
Added CLI wrapper

Yureien:
Tenacity can now be used as a CLI app to retry shell scripts or apps!

Closes https://github.com/jd/tenacity/issues/279

It's not yet complete, some of the advanced functions are left (such as exponential and random wait functions, thinking of a good way to implement those). A changelog is left too, but I wanted your reviews first on whether this is a good change, and if I am following the proper styling/coding guidelines or not.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,5 +8,6 @@ AUTHORS
 ChangeLog
 .eggs/
 doc/_build
+venv/
 
 /.pytest_cache
--- a/bin/tenacity
+++ b/bin/tenacity
@@ -1,0 +1,3 @@
+#!/usr/bin/env python3
+from tenacity.__main__ import main
+main()
--- a/setup.py
+++ b/setup.py
@@ -18,4 +18,5 @@ import setuptools
 setuptools.setup(
     setup_requires=["setuptools_scm"],
     use_scm_version=True,
+    scripts=["bin/tenacity"],
 )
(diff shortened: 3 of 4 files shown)


.gitignore
```diff
@@ -8,5 +8,6 @@ AUTHORS
 ChangeLog
 .eggs/
 doc/_build
+venv/
```
jd, line 11: not needed

bin/tenacity
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env python3
+from tenacity.__main__ import main
```
jd, line 2: use console scripts

setup.py
```diff
@@ -18,4 +18,5 @@
 setuptools.setup(
     setup_requires=["setuptools_scm"],
     use_scm_version=True,
+    scripts=["bin/tenacity"],
```
jd, line 21: use console scripts

jd (changes requested):
Not convinced about the need/usage, but why not. Though there's no test so I'm a bit reluctant to merge that.

tenacity/__main__.py
```diff
@@ -0,0 +1,108 @@
...
+    dest="wait_time",
+    type=int,
+    help="Fixed time to wait between each attempt in seconds.",
+)
+parser.add_argument(
+    "--ignore-code",
+    dest="ignore_codes",
+    action="append",
+    type=int,
+    help="Status code that should be ignored. This argument can be called multiple times.",
```
asqui, line 52:
```suggestion
    help="Non-zero status code that should be ignored. This argument can be called multiple times.",
```

asqui:
This seems like a nice feature!

Worth mentioning in the release notes, and would be good to have some test coverage.

jd:
Not updated in a while.

Issue #326, opened 2021-09-08
retry_if_result always use None as a callback parameter


Draqun:
Hi.
I couldn't find answer in documentation so I'm asking here. How to force attempt manager to put value returned by function into callback, when I iterate over Retrying. Example of code is below. In this example `is_not_expected` always has a None in value parameter. There is possibility to pass result variable content into `is_not_expected` via attempt/retryer?

```python
from tenacity import Retrying, retry_if_result, stop_after_attempt, wait_fixed


def get_string():
    return 'NOT-TEST'


def is_not_expected(value):
    return value != 'TEST'


def main():
    retryer = Retrying(
        retry=retry_if_result(is_not_expected),
        stop=stop_after_attempt(3),
        wait=wait_fixed(0),
        reraise=True,
    )
    for attempt in retryer:
        with attempt:
            result = get_string()
    return result

main()

```

Draqun:
Solution

```python
from tenacity import Retrying, retry_if_result, stop_after_attempt, wait_fixed


def get_string():
    return 'NOT-TEST'


def is_not_expected(value):
    return value != 'TEST'


def main():
    retryer = Retrying(
        retry=retry_if_result(is_not_expected),
        stop=stop_after_attempt(3),
        wait=wait_fixed(0),
        reraise=True,
    )
    for attempt in retryer:
        with attempt:
            result = get_string()
        attempt.retry_state.set_result(result)  # SOLUTION
    return result

main()
```

Please add this example to documentation and I think it can be close.

jd:
When using `Retrying` as an iterator, `retry_if_result` doesn't see the result because tenacity isn't calling the function — you are. The simplest fix is to use the decorator style instead:

```python
@retry(retry=retry_if_result(is_not_expected), stop=stop_after_attempt(3), wait=wait_fixed(0))
def get_string():
    return 'NOT-TEST'
```

Closing as a usage question.

PR #369, opened 2022-10-26
Fix async loop with retrying code block when result is available

victormferrara:
Hi!

I was facing an issue of not being able to receive the result payload in callbacks when using the retrying code block pattern. It's basically what's explained in https://github.com/jd/tenacity/issues/326, which has a workaround.

However, when using an async loop it was having a strange behaviour when the block was finally successful. The minimal example is present in the test:
```python
async def test():
    attempts = 0
    async for attempt in tasyncio.AsyncRetrying(retry=retry_if_result(lambda x: x < 3)):
        with attempt:
            attempts += 1
        attempt.retry_state.set_result(attempts)
    return attempts

result = await test()
```
Without the fix, the last iteration of the loop receives the result instead of an `AttemptManager` and it crashes in the `with` clause.

Cheers!


victormferrara:
@jd Sorry to bother. Any thoughts on this PR? Thanks!

jd (approved):

PR #379, opened 2023-01-30
ci: add github action jobs

sileht:


jd:
I guess you want to change Mergify configuration too?

sileht:
> I guess you want to change Mergify configuration too?

Because of the matrix, I will know the name once they run at least once.

jd (approved):
Merging and praying it triggers at some point.

PR #382, opened 2023-01-30
ci: add mypy

sileht:
This a rework of https://github.com/jd/tenacity/pull/339.

The main differences with the PR above:
* I set the ignore code for all `# type: ignore`
* Instead of taking decision when RetryCallState has some attributes not
  set as expected, I raised a RuntimeError as this is unexpected.
  Deciding on such a situation would just hide bugs.
* Add strict=True


jd (approved):

jd (dismissed):

jd (approved):

sileht:
@mergifyio update

PR #384, opened 2023-01-30
fix(mypy): add missing typing

### LINKED ISSUE #324: Using stop, wait and retry custom callbacks with Retrying causes mypy failures
I am using custom callbacks for `stop` and `wait` in the form of functions accepting a retry state, passing them to the `Retrying` context manager, as documented in [the Custom Callbacks section](https://tenacity.readthedocs.io/en/latest/#other-custom-callbacks). With the latest version of tenacity, this is now causing `mypy` failures since `BaseRetrying` [now only accepts `wait` and `stop` callbacks](https://github.com/jd/tenacity/blob/a7f548520e4ee871ad8aeb354ecfa1a324c8ca19/tenacity/__init__.py#L228-L230) that are function objects of a class inheriting from `wait_base` and `stop_base`. The same will also happen if a `retry` custom callback was provided.

These changes were introduced in #304. 

Since wait_base and stop_base aren't exposed as part of the public API, I believe this should be updated in order to enable using `Retrying` directly with `stop`, `wait` and `retry` custom callbacks as documented.

Small example causing `mypy` failures:
``` python
import tenacity

def my_stop(retry_state: tenacity.RetryCallState) -> bool:
    return True


def my_retry(retry_state: tenacity.RetryCallState) -> bool:
    return False


def my_wait(retry_state: tenacity.RetryCallState) -> bool:
    return False


for attempt in tenacity.Retrying(retry=my_retry, wait=my_wait, stop=my_stop):
    with attempt:
        pass
```

`mypy` output:

``` shell
$ python3.9 -m mypy test.py 
test.py:15: error: Argument "retry" to "Retrying" has incompatible type "Callable[[RetryCallState], bool]"; expected "retry_base"
test.py:15: error: Argument "wait" to "Retrying" has incompatible type "Callable[[RetryCallState], bool]"; expected "wait_base"
test.py:15: error: Argument "stop" to "Retrying" has incompatible type "Callable[[RetryCallState], bool]"; expected "stop_base"
Found 3 errors in 1 file (checked 1 source file)
```

If we change the same example to use the `@retry` decorator instead of the `Retrying` context manager, `mypy` reports no errors:

``` python
import tenacity

def my_stop(retry_state: tenacity.RetryCallState) -> bool:
    return True


def my_retry(retry_state: tenacity.RetryCallState) -> bool:
    return False


def my_wait(retry_state: tenacity.RetryCallState) -> bool:
    return False

@tenacity.retry(wait=my_wait, stop=my_stop, retry=my_retry)
def retryable_function():
    pass

retryable_function()
```

`mypy` output:
```
$ python3.9 -m mypy test.py 
Success: no issues found in 1 source file
```

sileht:
Runtime callbacks for retry/stop/wait allow to pass a function or a
class that inherit from retry_base/wait_base/stop_base.

This fixes the typing when a function is used.

Fixes #324

### DIFF
--- a/tenacity/__init__.py
+++ b/tenacity/__init__.py
@@ -86,8 +86,9 @@ except ImportError:
 if t.TYPE_CHECKING:
     import types
 
-    from .wait import wait_base
-    from .stop import stop_base
+    from .retry import RetryBaseT
+    from .stop import StopBaseT
+    from .wait import WaitBaseT
 
 
 WrappedFn = t.TypeVar("WrappedFn", bound=t.Callable[..., t.Any])
@@ -217,9 +218,9 @@ class BaseRetrying(ABC):
     def __init__(
         self,
         sleep: t.Callable[[t.Union[int, float]], None] = sleep,
-        stop: "stop_base" = stop_never,
-        wait: "wait_base" = wait_none(),
-        retry: retry_base = retry_if_exception_type(),
+        stop: "StopBaseT" = stop_never,
+        wait: "WaitBaseT" = wait_none(),
+        retry: "RetryBaseT" = retry_if_exception_type(),
         before: t.Callable[["RetryCallState"], None] = before_nothing,
         after: t.Callable[["RetryCallState"], None] = after_nothing,
         before_sleep: t.Optional[t.Callable[["RetryCallState"], None]] = None,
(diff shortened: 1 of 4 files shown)


jd (approved):

sileht:
@mergifyio update

PR #386, opened 2023-01-31
fix: remove __iter__ from AsyncRetrying

sileht:
This method doesn't make sense for AsyncRetrying. And it's even more
vicious as:

for attempts in AsyncRetrying():
    with attempts:
        await do_for_a_while()

will work, but sleep(XXX) corotine will never be awaited and the call is
retried instantly.

This change is kind of breaking change as BaseRetrying interface change
and a new interface BaseSyncRetrying.

Another alternative is to raise an RuntimeError in AsyncRetrying if
__iter__ is used.


sileht:
I prefer this version: https://github.com/jd/tenacity/pull/387

PR #387, opened 2023-01-31
fix: remove __iter__ from AsyncRetring

sileht:
This method doesn't make sense for AsyncRetrying. And it's even more
vicious as:

```
for attempts in AsyncRetrying():
    with attempts:
        await do_for_a_while()
```

will work, but sleep(XXX) coroutine will never be awaited and the call is
retried instantly.

This change removes the __iter__ method from AsyncRetrying.

### DIFF
--- a/releasenotes/notes/no-async-iter-6132a42e52348a75.yaml
+++ b/releasenotes/notes/no-async-iter-6132a42e52348a75.yaml
@@ -1,0 +1,7 @@
+---
+fixes:
+  - |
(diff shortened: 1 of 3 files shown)


jd (changes requested):
Can you add a test for this?

jd (approved):

jd:
@sileht can you add a release note?

jd (approved):

sileht:
@mergifyio update

jd (approved):

Issue #399, opened 2023-04-14
Add type overloads for async `sleep` function


Zac-HD:
```python
import asyncio  # or Trio, etc.
import tenacity

@tenacity.retry(
    sleep=asyncio.sleep,  # typecheckers will error here!
)
async def foo():
    pass
```
The type annotation here
https://github.com/jd/tenacity/blob/433324956abb028f6d993195d31e4dd8308115c3/tenacity/__init__.py#L500-L502
should be `t.Callable[[t.Union[int, float]], t.Union[None, t.Awaitable[None]]`, since `AsyncRetrying` allows async `sleep` functions.

Going further, we could define an additional `AsyncWrappedFn` typevar and propagate that though to _ensure_ that the `sleep` callback is always awaitable when retrying for a coroutine function.  This is a pretty important performance footgun to avoid!

PR #400, opened 2023-04-18
Add support for async sleep functions in tenacity.retry annotation.

jakkdl:
Fixes https://github.com/jd/tenacity/issues/399

Since tests aren't typechecked I didn't add a test for this, although if you want I could open another PR with tox modifications and `# mypy: disable-error-codes= I added to check it locally.

### DIFF
--- a/tenacity/__init__.py
+++ b/tenacity/__init__.py
@@ -499,7 +499,7 @@ def retry(func: WrappedFn) -> WrappedFn:
 
 @t.overload
 def retry(
-    sleep: t.Callable[[t.Union[int, float]], None] = sleep,
+    sleep: t.Callable[[t.Union[int, float]], t.Optional[t.Awaitable[None]]] = sleep,
     stop: "StopBaseT" = stop_never,
     wait: "WaitBaseT" = wait_none(),
     retry: "RetryBaseT" = retry_if_exception_type(),


tenacity/__init__.py
```diff
@@ -499,7 +499,7 @@ def retry(func: WrappedFn) -> WrappedFn:
 
 @t.overload
 def retry(
-    sleep: t.Callable[[t.Union[int, float]], None] = sleep,
+    sleep: t.Callable[[t.Union[int, float]], t.Union[t.Awaitable[None], None]] = sleep,
```
jd, line 502:
```suggestion
    sleep: t.Callable[[t.Union[int, float]], t.Optional[t.Awaitable[None]] = sleep,
```

jd:
We have mypy running but it's possible that's missing some tests, if you have some that'd be great.

tenacity/__init__.py
jakkdl: oh yeah of course, I've gotten too used to `from __future__ import annotations` and just doing `| None`

Zac-HD:
@jd any timeline on this, or things I can do to help?  It's currently pretty painful to use tenacity in async codebases and I'd love to get the patch shipped asap 🙏

jakkdl:
@jd any blocker for going ahead and merging this?

jakkdl:
Thanks!

PR #401, opened 2023-04-18
Typecheck tests

jakkdl:
Runs the `tests/` folder against mypy as well. Which I've found to be a great way to catch annotation errors in typechecked code in my projects, but I didn't bother trying to typehint everything and just threw up enough ignore/disables until mypy got silent.

Forked on top of #400, adding a test for it.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -23,7 +23,7 @@ deps = flake8
        flake8-docstrings
        flake8-rst-docstrings
        flake8-logging-format
-commands = flake8
+commands = flake8 {posargs}
 
 [testenv:black]
 deps =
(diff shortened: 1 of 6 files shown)


tox.ini
```diff
@@ -34,8 +34,9 @@ commands =
 [testenv:mypy]
 deps =
     mypy>=1.0.0
+    pytest # for stubs
 commands =
-    mypy tenacity
+    mypy tenacity tests
```
jd, line 39:
```suggestion
    mypy
```
`files` has been modified so I guess it should work?

jakkdl: yepyep, fixed

jakkdl:
also added `{posargs}` to mypy & flake8, was handy to be able to check that mypy checked the correct files by passing it `-v`

jd (approved):

PR #406, opened 2023-07-06
Preserve function default and kwdefault through retry decorator

michael-motional:
Hello, and thanks for the great package. I noticed that the `retry` decorator uses `functools.wraps`. By default, this means that `__defaults__` and `__kwdefaults__` will always be `None` on the wrapped function. In general this is reasonable behavior, since a wrapper could change these defaults. However, I believe the retry decorator does not, and thus should preserve these attributes. I've created a patch to change this, please let me know what you think :)


michael-motional:
@jd When you have some time, could I get a review on this? :pray:

jd (approved):

jd (approved):

PR #408, opened 2023-07-10
Improved View for python snippet

Mgajurel:
Python snippet were not being displayed as python codes, so fixed Sphinx directives to display python snippet.


jd (commented):
AFAIU this breaks the sphinx module that test the example though?

Mgajurel:
Yeah, I figured that. How about we create a different Readme.md here for the Repo?

Mgajurel:
Writing a script to convert from rst to md and have a GitHub action to automate the process, would even be better or would it be an overkill? @jd

Mgajurel:
Should I close this PR and may be create a new one? @jd

PR #419, opened 2023-09-27
Add `RetryCallState` to the API docs

rouge8:
User code can access the `RetryCallState` so its members should be documented.

### DIFF
--- a/doc/source/api.rst
+++ b/doc/source/api.rst
@@ -17,6 +17,9 @@ Retry Main API
 .. autoclass:: tenacity.tornadoweb.TornadoRetrying
    :members:
 
+.. autoclass:: tenacity.RetryCallState
+   :members:
+
 After Functions
 ---------------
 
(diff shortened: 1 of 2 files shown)


jd (approved):

aragaer:
Current code assumes that there is always a function name available (there even was a comment saying that this should never happen) -- which is not the case when using a context manager. Currently logging functions would output "Retrying <unknown>" and there is no way to specify which section of code failed and needs the retry. The idea is to specify an optional label for a retry block. And if retry block is repeated in a loop it also allows to specify which iteration caused a retry:
```python
for item in some_iterable:
    try:
        for attempt in Retrying(
                label=f"doing stuff with {item}",
                stop=stop_after_attempt(5),
                before_sleep=before_sleep_log(logger, logging.DEBUG),
                wait=wait_fixed(10)):
            with attempt:
                data = do_stuff(item)
    except RetryError:
        continue
    ...
```

rouge8:
Oh, It's actually documented in the README...

rouge8:
It renders correctly in Sphinx locally (using the Alabaster theme) after running `tox -e py311`, so I'm not sure what's going on there.

jd (approved):

PR #424, opened 2023-12-03
typing: update version compare to support pyright/pylance

sbdchd:
The current version comparison works with mypy but pyright doesn't handle it.

Comparing against minor and major version allows it to work with pyright.

related: https://github.com/microsoft/pyright/issues/6639


sbdchd:
probably doesn't need a change log entry?

jd (approved):

PR #426, opened 2023-12-18
Improves wait_exponential_jitter

yxtay:
# Summary

Improves wait_exponential_jitter to inherit wait_exponential
    - Reuses `wait_exponential.__call__` method instead of of duplicating code
    - Follows argument names of wait_exponential
    - Adds `min` argument
    - Supports supplying `max`, `jitter`, `min` arguments as `timedelta`

# Tests

```
pytest tests/test_tenacity.py::TestWaitConditions::test_wait_exponential_jitter
```

### DIFF
--- a/releasenotes/notes/improve-wait-exponential-jitter-e0a3a27fef39f97f.yaml
+++ b/releasenotes/notes/improve-wait-exponential-jitter-e0a3a27fef39f97f.yaml
@@ -4,5 +4,5 @@ features:
     Improves wait_exponential_jitter to inherit wait_exponential
     - Reuses `wait_exponential.__call__` method instead of reimplementing from scratch
     - Follows argument names of wait_exponential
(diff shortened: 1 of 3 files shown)


jd (approved):

yxtay:
Is there anything blocking this from being merged? Deprecation concerns?

krispenney:
:wave: I was just looking at implementing the `min` and the typing improvements that you added here @yxtay, I feel like they'd still be nice api improvements. 

Would it make sense to split those commits off to be merged while @LotfiRafik 's comments are addressed?

jd:
Superseded by #628 and #629, which implement the features from this PR (min param, timedelta support, multiplier) while avoiding the inheritance-based approach that introduced the clamping-order bug. Thanks for the ideas @LotfiRafik!

PR #289, opened 2021-04-13
Add support for an async defined retry and callback_error_retry

symphony-youness:
This issue raises the need for an asynchronously defined error callback: #249 

This is a first approach to be able to use an asynchronously defined retry (on every retry iteration) and callback_error_retry (after the retrying is stopped)

- I added tests for both
- I added an await in __call__ and __anext__ to be able to use the context manager (passed async context manager tests)
- I had to override the iter() in AsyncRetrying to define it as async

PR #433, opened 2024-01-29
Add support for async strategies and callbacks

### LINKED ISSUE #249: Is an async retry_error_callback possible?

hasier:
Fixes #249 
Follows up on #289 #363 

I took a stab at [DRYing the `iter()` function](https://github.com/jd/tenacity/pull/363#pullrequestreview-1118442473) in order to support `async` callbacks, and once those were possible, I thought it would also make sense to allow for `async` strategies.

The approach I have taken is to build a list of actions that `iter()` needs to go through, where each action is just a piece of the whole current logic, and each step can further extend the list of actions. 3 pieces have been taken out to redefine in specific Retrying implementations, i.e. `asyncio`. This way we can make sure we use coroutines for those calls, as well as dynamically wrapping each function as a coroutine, allowing for both sync and async strategies and callbacks.

I tried to make somewhat meaningful commits, it should be easier to review commit-by-commit.

I appreciate the new approach looks quite odd and is not very intuitive to follow, happy to take some feedback and improve it if anyone has some other ideas.

### DIFF
--- a/tenacity/__init__.py
+++ b/tenacity/__init__.py
@@ -312,9 +312,8 @@ class BaseRetrying(ABC):
         self.statistics["attempt_number"] = 1
         self.statistics["idle_for"] = 0
 
-    @classmethod
-    def _wrap_action_func(cls, fn: t.Callable[..., t.Any]) -> t.Callable[..., t.Any]:
-        return fn
+    def _add_action_func(self, fn: t.Callable[..., t.Any]) -> None:
+        self.iter_state["actions"].append(fn)
 
     def _run_retry(self, retry_state: "RetryCallState") -> None:
         self.iter_state["retry_run_result"] = self.retry(retry_state)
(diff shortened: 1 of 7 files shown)


jd:
Is there a way to split that in smaller changes? It's a bit large to review as is, even if it looks great

hasier:
> Is there a way to split that in smaller changes? It's a bit large to review as is, even if it looks great

@jd fair point. I have split it into 3 PRs, hopefully this works for you.
1. https://github.com/jd/tenacity/pull/434
2. https://github.com/hasier/tenacity/pull/1 stacked on the previous one, but in my repo. Let me know if you prefer I did it against your `main` instead.
3. https://github.com/hasier/tenacity/pull/2 stacked on the previous one, but in my repo too. Let me know if you prefer I did it against your `main` instead.

PR #434, opened 2024-02-02
Incrementally build iter actions list

hasier:
First PR after breaking down https://github.com/jd/tenacity/pull/433
- Second PR https://github.com/hasier/tenacity/pull/1
- Third PR https://github.com/hasier/tenacity/pull/2

An attempt at [DRYing the `iter()` function](https://github.com/jd/tenacity/pull/363#pullrequestreview-1118442473) in order to support `async` callbacks.

The approach I have taken is to build a list of actions that `iter()` needs to go through, where each action is just a piece of the whole current logic, and each step can further extend the list of actions. 3 pieces have been taken out to redefine in specific Retrying implementations, i.e. `asyncio`. This way we can make sure we use coroutines for those calls, as well as dynamically wrapping each function as a coroutine, allowing for both sync and async strategies and callbacks.

### DIFF
--- a/tenacity/__init__.py
+++ b/tenacity/__init__.py
@@ -279,6 +279,14 @@ class BaseRetrying(ABC):
             self._local.statistics = t.cast(t.Dict[str, t.Any], {})
             return self._local.statistics
 
+    @property
+    def iter_state(self) -> t.Dict[str, t.Any]:
+        try:
+            return self._local.iter_state  # type: ignore[no-any-return]
+        except AttributeError:
+            self._local.iter_state = t.cast(t.Dict[str, t.Any], {})
+            return self._local.iter_state
+
     def wraps(self, f: WrappedFn) -> WrappedFn:
         """Wrap a function for retrying.
 
@@ -303,20 +311,13 @@ class BaseRetrying(ABC):
         self.statistics["attempt_number"] = 1
         self.statistics["idle_for"] = 0
 
-    def iter(self, retry_state: "RetryCallState") -> t.Union[DoAttempt, DoSleep, t.Any]:  # noqa
-        fut = retry_state.outcome
-        if fut is None:
-            if self.before is not None:
-                self.before(retry_state)
-            return DoAttempt()
+    def _add_action_func(self, fn: t.Callable[..., t.Any]) -> None:
+        self.iter_state["actions"].append(fn)
 
-        is_explicit_retry = fut.failed and isinstance(fut.exception(), TryAgain)
-        if not (is_explicit_retry or self.retry(retry_state)):
-            return fut.result()
+    def _run_retry(self, retry_state: "RetryCallState") -> None:
+        self.iter_state["retry_run_result"] = self.retry(retry_state)
 
-        if self.after is not None:
-            self.after(retry_state)
-
+    def _run_wait(self, retry_state: "RetryCallState") -> None:
         if self.wait:
             sleep = self.wait(retry_state)
         else:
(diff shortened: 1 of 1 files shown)


tenacity/__init__.py
```diff
@@ -279,6 +279,14 @@ def statistics(self) -> t.Dict[str, t.Any]:
             self._local.statistics = t.cast(t.Dict[str, t.Any], {})
             return self._local.statistics
 
+    @property
+    def iter_state(self) -> t.Dict[str, t.Any]:
+        try:
+            return self._local.iter_state  # type: ignore[no-any-return]
+        except AttributeError:
+            self._local.iter_state = t.cast(t.Dict[str, t.Any], {})
```
jd, line 287: Can we have something either more structured or split?
I am no fan of having a dict with `Any` and random keys TBH.

At least a `TypedDict`?

hasier: Sounds fair, wdyt? https://github.com/jd/tenacity/pull/434/commits/538686b75bd90f15c6c7ead7999904620fa1f910

tbh I just copied the approach from the `statistics` property, and I was tempted to change the `local()` implementation for a `ContextVar` and maybe introducing `dataclass`es, but seemed like a bit too much to do together with these changes. Maybe something to tackle separately if that's something you'd like to see.

hasier:
@jd I noticed `black` checks are also failing in CI, but I _think_ there is an incompatibility between `black` and `pep8`. The first one is trying to remove some whitespace, whereas the latter complains it wants the whitespace back. Is this something you can look at in the `main` branch maybe?

jd:
I switched the project to ruff to solve this.

hasier:
> I switched the project to ruff to solve this.

Thanks @jd! Just re-formatted the changes with `ruff` 😊

tenacity/__init__.py
```diff
@@ -97,6 +97,14 @@
 WrappedFn = t.TypeVar("WrappedFn", bound=t.Callable[..., t.Any])
 
 
+class IterState(t.TypedDict):
```
jd, line 100: using dataclass here?

tenacity/__init__.py
```diff
@@ -287,6 +295,14 @@ def statistics(self) -> t.Dict[str, t.Any]:
             self._local.statistics = t.cast(t.Dict[str, t.Any], {})
             return self._local.statistics
 
+    @property
+    def iter_state(self) -> IterState:
+        try:
+            return self._local.iter_state  # type: ignore[no-any-return]
+        except AttributeError:
+            self._local.iter_state = t.cast(IterState, {})
```
jd, line 303: …and avoid a cast here?

tenacity/__init__.py
hasier: Sounds good! https://github.com/jd/tenacity/pull/434/commits/dfd09074e62684e376264a5eb548645759797edd

tenacity/__init__.py
hasier: https://github.com/jd/tenacity/pull/434/commits/dfd09074e62684e376264a5eb548645759797edd

jd (approved):

hasier:
Sorry @jd `dataclass(slots=True)` is only available from Python 3.10 onwards 🤦 I just fixed it by conditionally adding it, hopefully that looks ok to you! https://github.com/jd/tenacity/pull/434/commits/5b11340f997b8db6e48728006bd0dc6257c446fe

hasier:
@jd how would you want to go about the chain of PRs? Do you want me to add a `reno` note here and merge this set of changes first, or would you prefer to merge the 2 other PRs here, then just merge 1 big PR to `main`? Happy to follow whichever process you prefer, let me know if you want me to update the PRs in any specific way.

jd (approved):

Issue #412, opened 2023-08-07
Passing an async function to before/after handler?


seidnerj:
I have the following code:

@retry(retry=retry_if_exception_type(SomeException),
       stop=stop_after_attempt(5),
       before_sleep=before_sleep_log(logging.getLogger(), logging.ERROR),
       wait=wait_exponential(multiplier=1, min=1, max=10),
       before=before_retry_handler)
async def async_wrapper(param):
    ....

async def before_retry_handler(retry_state):
    ....

Trouble is before_retry_handler has to be async because it calls async functions.

When I try to use the above, I get "RuntimeWarning: coroutine 'before_retry_handler' was never awaited self.before(retry_state)".

Am I missing something?

Thanks in advance

Hanabiraa:
I have not found a solution so far, so I made my own decorator. Before contributing to tenacity or my fork, for now I offer this solution (it can be optimized for your tasks):

```python3
import logging
from logging import Logger, getLogger
from typing import Callable, Coroutine, Type


class RetryError(BaseException):
    ...


def retry(
    after_callback: Callable[..., Coroutine] | None = None,
    logger: Logger | None = None,
    max_retries: int = 5,
    exception_for_retry: list[Type[Exception]] | Type[Exception] = Exception,
    reraise: bool = False,
):
    def func_wrapper(f):
        async def wrapper(*args, **kwargs):
            last_exc = None
            for attempt_num in range(max_retries):
                logger.info(msg=f"Trigger retrying. Attempt #{attempt_num}")
                try:
                    return await f(*args, **kwargs)
                except (
                    tuple(exception_for_retry)
                    if isinstance(exception_for_retry, list)
                    else exception_for_retry
                ) as exc:
                    last_exc = exc

                    if after_callback:
                        await after_callback()
            raise (last_exc if reraise else RetryError)

        return wrapper

    return func_wrapper
```

seidnerj:
That’s awesome, thanks a lot!

jd:
This already works with `AsyncRetrying`. All callbacks (`before`, `after`, `before_sleep`) are wrapped with `wrap_to_async_func` and properly awaited. Make sure you're using `AsyncRetrying` or that `@retry` is applied to an `async def` function (which automatically uses `AsyncRetrying` under the hood).

PR #437, opened 2024-02-06
Support async actions

### LINKED ISSUE #249: Is an async retry_error_callback possible?

hasier:
Fixes #249 #412 

Second PR after breaking down https://github.com/jd/tenacity/pull/433 (follows https://github.com/jd/tenacity/pull/434)
- First PR https://github.com/jd/tenacity/pull/434
- Third PR https://github.com/hasier/tenacity/pull/2

After [DRYing the `iter()` function](https://github.com/jd/tenacity/pull/434), make `AsyncRetrying` support `async` callbacks.

Supersedes https://github.com/hasier/tenacity/pull/1

### DIFF
--- a/tenacity/_asyncio.py
+++ b/tenacity/_asyncio.py
@@ -16,6 +16,7 @@
 # limitations under the License.
 
 import functools
+import inspect
 import sys
 import typing as t
 from asyncio import sleep
@@ -28,6 +29,15 @@ from tenacity import RetryCallState
 
 WrappedFnReturnT = t.TypeVar("WrappedFnReturnT")
 WrappedFn = t.TypeVar("WrappedFn", bound=t.Callable[..., t.Awaitable[t.Any]])
+
+
+def is_coroutine_callable(call: t.Callable[..., t.Any]) -> bool:
+    if inspect.isroutine(call):
+        return inspect.iscoroutinefunction(call)
+    if inspect.isclass(call):
+        return False
+    dunder_call = getattr(call, "__call__", None)  # noqa: B004
+    return inspect.iscoroutinefunction(dunder_call)
 
 
 class AsyncRetrying(BaseRetrying):
@@ -46,7 +56,7 @@ class AsyncRetrying(BaseRetrying):
 
         retry_state = RetryCallState(retry_object=self, fn=fn, args=args, kwargs=kwargs)
         while True:
-            do = self.iter(retry_state=retry_state)
+            do = await self.iter(retry_state=retry_state)
             if isinstance(do, DoAttempt):
                 try:
                     result = await fn(*args, **kwargs)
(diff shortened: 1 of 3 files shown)


tenacity/_asyncio.py
```diff
@@ -30,6 +31,15 @@
 WrappedFn = t.TypeVar("WrappedFn", bound=t.Callable[..., t.Awaitable[t.Any]])
 
 
+def is_coroutine_callable(call: t.Callable[..., t.Any]) -> bool:
```
jd, line 34: Could we have unit tests around this?

jd (changes requested):

tenacity/_asyncio.py
hasier: Good call! I ended up moving the function to `_utils` and slightly amending it (odd `partial` behaviour in 3.8 and 3.9, even though [docs say otherwise](https://docs.python.org/3/library/inspect.html#inspect.iscoroutinefunction)). wdyt? https://github.com/jd/tenacity/pull/437/commits/2616c4ad586f91b9cbad4e777f177fcc199911d2

tenacity/_asyncio.py
```diff
@@ -60,6 +61,47 @@ async def __call__(  # type: ignore[override]
...
+        if _utils.is_coroutine_callable(fn):
+            return fn
+
+        async def inner(*args: t.Any, **kwargs: t.Any) -> t.Any:
+            return fn(*args, **kwargs)
+
+        return inner
+
+    def _add_action_func(self, fn: t.Callable[..., t.Any]) -> None:
+        self.iter_state.actions.append(self._wrap_action_func(fn))
```
jd, line 75: Looks like this is the same as the one from `BaseRetrying`, no?

hasier: This one is slightly different, as it wraps the `fn` function with `self._wrap_action_func()`, which makes sure all calls are async.

I considered adding a separate `_wrap_action_func()` in `BaseRetrying` which could just `return fn` and override it for `asyncio`, to keep the `append()` logic just in `BaseRetrying`, but seemed a bit of an overkill considering it's just a one liner. Would love to hear your thoughts on it though.

hasier:
@jd any more thoughts on this? Happy to clarify anything else I may have missed.

tenacity/_asyncio.py
jd: Sounds good enough.

jd (approved):

hasier:
@jd I merged changes from `main` and tests seem to pass, feel free to re-approve and merge whenever you prefer 🙂

jd (approved):

PR #439, opened 2024-02-08
Add a Dependabot config to autoupdate GitHub action versions

kurtmckee:
This PR introduces a Dependabot configuration to auto-update GitHub action versions.

CI jobs are currently throwing deprecation warnings due to old versions of actions using Node 16 [[recent example](https://github.com/jd/tenacity/actions/runs/7798746570)]. If this PR merges, you can expect Dependabot to open a PR to update action versions as needed.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: 'github-actions'
+    directory: '/'
+    schedule:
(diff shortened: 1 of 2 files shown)


kurtmckee:
I think I've added an appropriate changelog fragment, but please let me know if anything else is needed. Thanks!

jd (approved):

PR #440, opened 2024-02-08
Add support for mocking sleep when retrying async functions

### LINKED ISSUE #360: Cannot easily mock out async sleep function
Issue #228 pointed out that it was difficult to patch the sleep function in tests. It was then fixed and now mocking `time.sleep` is straightforward.

Could the same be done for the async version `AsyncRetrying`? It would be nice if it was equally easy to mock `asyncio.sleep`.

snorfalorpagus:
This PR makes it easier to mock sleeping when retrying asynchronous functions, using the same approach as was done for synchronous functions in #236.

Example usage:

```python
@retry(wait=wait_fixed(9999999))
def function_with_retry():
    if random.random() < 0.9:
        raise RuntimeError("boom")
    return True

async def test_something(self):
    with unittest.mock.patch("tenacity.nap.asyncio_sleep", AsyncMock()):
        await function_with_retry()
```

Closes #360.

### DIFF
--- a/tenacity/_asyncio.py
+++ b/tenacity/_asyncio.py
@@ -18,13 +18,13 @@
 import functools
 import sys
 import typing as t
-from asyncio import sleep
 
 from tenacity import AttemptManager
 from tenacity import BaseRetrying
 from tenacity import DoAttempt
 from tenacity import DoSleep
 from tenacity import RetryCallState
+from tenacity.nap import async_sleep
 
 WrappedFnReturnT = t.TypeVar("WrappedFnReturnT")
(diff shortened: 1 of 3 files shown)


jd (approved):

snorfalorpagus:
@jd Let me know if there is anything else required before this can be merged. :)

jd (approved):

aaronkavlie-wf:
It would be nice to see this get in soon.
We monkeypatch the `sleep` property on the `retry` decorator in our tests, however that approach broke in 8.4.2.

mtvx:
Any change for getting this in?

snorfalorpagus:
I don't have the energy to resolve conflicts on this as it's been sitting idle for so long without merge. :(

PR #451, opened 2024-03-18
Add async strategies

hasier:
Third and last PR after breaking down https://github.com/jd/tenacity/pull/433 (follows ~https://github.com/hasier/tenacity/pull/1~ https://github.com/jd/tenacity/pull/437)
- First PR https://github.com/jd/tenacity/pull/434
- Second PR ~https://github.com/hasier/tenacity/pull/1~ https://github.com/jd/tenacity/pull/437

After [DRYing the `iter()` function](https://github.com/jd/tenacity/pull/434) and making `AsyncRetrying` support `async` callbacks, add new `async` strategies (basically a copy-paste) and improve typing.

Supersedes https://github.com/hasier/tenacity/pull/2


tenacity/asyncio/retry.py
```diff
@@ -0,0 +1,115 @@
...
+        super().__init__(predicate)  # type: ignore[arg-type]
+
+
+class retry_if_result(async_predicate_mixin, _retry_if_result, async_retry_base):  # type: ignore[misc]
+    """Retries if the result verifies a predicate."""
+
+    def __init__(
+        self, predicate: typing.Callable[[typing.Any], typing.Awaitable[bool]]
+    ) -> None:
+        super().__init__(predicate)  # type: ignore[arg-type]
```
jd, line 85: Is this really needed?

tenacity/retry.py
```diff
@@ -30,10 +30,16 @@ def __call__(self, retry_state: "RetryCallState") -> bool:
...
+
+    def __rand__(self, other: "retry_base") -> "retry_all":
+        return retry_all(other, self)
 
     def __or__(self, other: "retry_base") -> "retry_any":
-        return retry_any(self, other)
+        return other.__ror__(self)
+
+    def __ror__(self, other: "retry_base") -> "retry_any":
+        return retry_any(other, self)
```
jd, line 43: Could/should be part of a different PR AFAIU.

tenacity/asyncio/retry.py
hasier: Not strictly, but it helps with the typing, otherwise `mypy` and similar tools would complain about async functions.

tenacity/retry.py
hasier: Not really, as we are introducing new async strategies, if sync and async strategies got mixed then the process would fail, as async strategies [would not be awaited](https://github.com/jd/tenacity/blob/cb15300d9b4358d9bdb6cfd33d433368700b3abe/tenacity/retry.py#L266). This is the bit I had to introduce to allow that mix, without this [these tests don't pass](https://github.com/jd/tenacity/pull/451/files#diff-55234053c6f9538eec8b76e6fce1715585c5359a3a8362042d1041381c6bd751R258-R364).

jd: I'm not saying it should be merged after, but before as it could be useful for even current tenacity code, no?

hasier: Ah I see. What would you like to see then, just these changes for `__rand__` and `__ror__` in a separate PR? I can do that, it just didn't feel that useful because as the codebase is now, there's only ever 1 type of `retry_base` and no `async` calls anywhere, so there won't be any undesired mix that makes these operations useful. It's in this PR that I'm introducing new `async` strategies that need these to avoid errors when mixing both strategies. Or maybe I am missing your point entirely and it's something else you are expecting 😅

hasier: @jd any thoughts on my latest comment? 😊

tenacity/asyncio/retry.py
jd: Right, that does not look good though.
I think `self.predicate` is going to have different types than the one defined in `retry_base`. I think we might need to not use `retry_base` or simply use a different async base.

tenacity/retry.py
jd: Reading the tests you linked I'm wondering how it even works since there's no `await` being used by `retry_any`. What do I miss?

jd:
@hasier sorry for the long delay, that PR slipped my mind. Don't hesitate to harass me more for reviews.

tenacity/retry.py
hasier: [The async `retry_any` does use `await`](https://github.com/jd/tenacity/pull/451/files#diff-e40ea0a413e972c47c884a384af80c3cbe4934358aa1d330447905320c8177d2R88-R100). This in combination with the `__ror__` and `__rand__` in the base class makes sure that we always use the appropriate any/all version.
- If we do `sync_strategy | sync_strategy` then it uses the regular `__and__` for the first strategy, so it results in using the sync `retry_any`.
- If we do `async_strategy | sync_strategy` then it uses the regular `__and__` for the first strategy, so it results in using the async `retry_any`.
- If we do `sync_strategy | async_strategy` then it uses the `__rand__` for the second strategy, as it's a subclass of the first one, so it results in using the async `retry_any`.
- If we do `async_strategy | async_strategy` then it uses the regular `__and__` for the first strategy, so it results in using the async `retry_any`.

I agree it does feel a bit odd, but I think this results in quite a good abstraction, as the setup in the base `retry_any`/`retry_all` allows other types of strategies (in the case of the changes here, async bases) to override this behaviour, while the original sync bases don't need to know about these implementations at all.

The tests I linked work because regardless of the types of strategies used, they are always triggered when using `AsyncRetrying(retry=retry_strategy)`, which requires doing so with `async for`, it's then the internals that make sure all of the callables and strategies are async and called as such.

tenacity/asyncio/retry.py
hasier: That's fair. I mainly wanted to keep it dry and reuse as much as possible. Would you rather have a whole new (mainly copy-paste) async implementation for these cases then?

hasier:
@jd sorry, I've been busy these last few days, could you take another look at [my latest comment](https://github.com/jd/tenacity/pull/451#discussion_r1593883436)? 🙏

tenacity/asyncio/retry.py
jd: I think it'd be better. As we started to type everything, I know from experience that trying to be smarter than mypy ends up haunting you at some point.

hasier: hahah fair, done! https://github.com/jd/tenacity/pull/451/commits/927965d09d4091c7a72fee896a7644fe4f5d2d78

hasier:
@jd any idea what might be going on with the Tornado tests? https://github.com/jd/tenacity/actions/runs/9347251510/job/25723863307?pr=451

hasier:
@jd seems like [a fix has already been merged and released for the Tornado issue](https://github.com/tornadoweb/tornado/pull/3382). I guess that means you can close https://github.com/jd/tenacity/issues/460, as [tests should automatically pick the latest version](https://github.com/jd/tenacity/blob/cb15300d9b4358d9bdb6cfd33d433368700b3abe/setup.cfg#L40). Could you rerun the checks here as well, please? 🙏

jd:
@mergify update

jd (dismissed):

jd:
@hasier can we add a release note?

hasier:
> Add release note

Good point @jd, done! https://github.com/jd/tenacity/pull/451/commits/4aa9ccf271cb37b0ac37437075bb25b033a2da0a

Also, I just realised finally I only added a [base](https://github.com/jd/tenacity/pull/451/files#diff-e40ea0a413e972c47c884a384af80c3cbe4934358aa1d330447905320c8177d2R26-R51) for `retry`, but nothing for `stop` or `wait`. Would you want me to replicate this together with the matching `all`/`any` to allow for easier extension of those in the future? Not sure if there might be similar use cases for those as I don't see them using any functions as predicates, but asking just in case.

hasier:
Thanks for the review and merging @jd! 🙇 Are you expecting to release a new version soon? 😁

Knamdev:
asyncio folder is missing under tenacity when downloaded

PR #463, opened 2024-05-24
Support Trio out-of-the-box, take 2

jakkdl:
I didn't have push rights to #295, so opening a new PR from a new fork. This is just taking @njsmith's PR, merging main, fixing conflicts and making it work with the new code.


Zac-HD:
@jd anything I can do to help with this?  It'd be really nice to delete all the `sleep=trio.sleep` calls at work 😅

doc/source/index.rst
```diff
@@ -568,28 +568,34 @@ in retry strategies like ``retry_if_result``. This can be done accessing the
...
 
     @retry
-    async def my_async_function(loop):
+    async def my_asyncio_function(loop):
         await loop.getaddrinfo('8.8.8.8', 53)
 
+.. code-block:: python
+
+    @retry
+    def my_async_trio_function():
```
jd, line 583:
```suggestion
    async def my_async_trio_function():
```
no?

tenacity/_asyncio.py
```diff
@@ -30,19 +30,31 @@
...
+    # If trio isn't already imported, then it's definitely not running, so we
+    # can skip further checks.
+    if "trio" in sys.modules:
+        # If trio is available, then sniffio is too
+        import trio
+        import sniffio
+
+        if sniffio.current_async_library() == "trio":
+            await trio.sleep(seconds)
+            return
```
jd, line 44:
```suggestion
    if "sniffio" in sys.modules:
        import sniffio

        if sniffio.current_async_library() == "trio":
            import trio
            await trio.sleep(seconds)
            return
```

Sounds slightly better?

tenacity/_asyncio.py
```diff
@@ -30,19 +30,31 @@
...
+    # Otherwise, assume asyncio
     # Lazy import asyncio as it's expensive (responsible for 25-50% of total import overhead).
     import asyncio
 
-    return asyncio.sleep(duration)
+    await asyncio.sleep(seconds)
```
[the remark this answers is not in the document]
jakkdl: The reason was is because that's how #295 did it and I didn't think about it when merging it with main :sweat_smile: 
Thinking about it now, doing it like #450 would be more performant (skipping `sniffie.current_async_library` check whenever calling sleep) - but it could maybe break if you're decorating an anyio test function that is parametrized against different async backends? Not 100% sure, but I can try to write a test that would trigger that.

tenacity/_asyncio.py
jakkdl: Yeah it does sound better, though I suppose it may be a slight performance loss. If sniffio is in `sys.modules` then importing it will be instant. [`sniffio.current_async_library()`](https://github.com/python-trio/sniffio/blob/a1cc1692f5be32645f65ebeb33e6a138180359e0/sniffio/_impl.py#L25) doesn't look super expensive, but if we're sticking with the status quo of this function in its entirety being called every time we want to sleep then it may be worth having slightly uglier code in exchange for a minor speedup.

doc/source/index.rst
jakkdl: yeah that looks like an obvious mistake from #295

tenacity/_asyncio.py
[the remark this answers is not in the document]
jakkdl: nowait I'm dumb, briefly forgot how async works. Yeah this makes no functional difference given the current code in `AsyncRetrying.__call__` where it immediately awaits the function after calling it. So it's just a question of whether one prefers an `async def` that itself calls `await` and returns `None` vs having a sync `def` that returns an `Awaitable[None]`. In general I find the former to be more readable and intuitive, but am fine with either.

tests/test_tornado.py
```diff
@@ -72,6 +72,10 @@ def retryable(thing):
         finally:
             gen.is_coroutine_function = old_attr
 
+    # temporary workaround for https://github.com/jd/tenacity/issues/460
+    def runTest(self):
+        self.subTest()
```
jd, line 77: should not be needed
```suggestion
```

tests/test_tornado.py
```diff
@@ -72,6 +72,8 @@ def retryable(thing):
         finally:
             gen.is_coroutine_function = old_attr
 
+    # temporary workaround for https://github.com/jd/tenacity/issues/460
+
```
jd, line 76:
```suggestion
```

tenacity/_asyncio.py
jd: Not sure I follow your reasoning here. In your code there are 2 imports each time whereas there's one or two in the code I suggested, so I'm not sure how my version is worse? 🤔

tenacity/_asyncio.py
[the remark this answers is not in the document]
jd: I'd prefer to not change things that don't need to be changed 😁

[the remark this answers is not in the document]
jakkdl: fixed :)

jmehnle:
Came here to watch the showdown. If this lands, it's gonna tip my scale in favor of `tenacity` over [`backoff`](https://github.com/litl/backoff), which still hasn't made any effort to support `trio`. Thanks for your efforts, gentlefolk!

Zac-HD:
> Does the current implementation of `_portable_async_sleep()` handle correctly the case where one is running on a `trio` event loop, but is using [`trio-asyncio`](https://github.com/python-trio/trio-asyncio) to run `asyncio` libraries inside the `trio` event loop?

Yes, [`sniffio` handles this for us](https://sniffio.readthedocs.io/en/latest/#usage) 🙂

jd (approved):

tenacity/_asyncio.py
jd: Well your version expect trio to be the backend just because it's imported. Not sure if it's better but I admit I don't know sniffio nor trio enough to judge 😆

PR #465, opened 2024-05-31
Update index.rst: Remove *

### LINKED ISSUE #462: Readme formatting fell apart
Screenshot:
image

TheRealBecks:
Fixes #462


jd (dismissed):

TheRealBecks:
@jd Do I need to do something here?

jd:
I don't think so, though we need to fix the CI, but that should be a different PR.

jd (approved):

Issue #471, opened 2024-06-17
ImportError tenacity.asyncio


vogre:
Using Python 3.11, and latest tenacity 8.4.0

```
> import tenacity
...
lib/python3.11/site-packages/tenacity/__init__.py", line 653, in <module>
     from tenacity.asyncio import AsyncRetrying  # noqa:E402,I100
     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 ModuleNotFoundError: No module named 'tenacity.asyncio'
```

GabLinares:
Facing the same issue, seems to be related to a change made in a [recent PR](https://github.com/jd/tenacity/pull/451)

Peelz:
Also getting this error, fixed by update requirements.txt `tenacity>=8.2.3,<8.4.0`

dpinol:
Strange that CI didn't catch it right?

pmercier:
Same here with python 3.12.
What is strange is that asyncio is present in the zip and tar.gz package, but abscent when installed with pip or poetry.
![image](https://github.com/jd/tenacity/assets/584193/c3fc1f78-eec6-45c2-b0b4-0ef7b8b26d50)
![image](https://github.com/jd/tenacity/assets/584193/a823dfd6-7180-4d36-9ee1-87cea32ae285)

JeromeDuboisVizzia:
Hey same error here in a AWS lambda using python 3.12, I didn't encounter it last week :-/
```
[ERROR] Runtime.ImportModuleError: Unable to import module 'app': No module named 'tenacity.asyncio'
Traceback (most recent call last):
```
Thanks @Peelz for the workaround

harupy:
Encountered the same issue. It looks like the `asyncio` module is missing:

```
% zipinfo -1 ~/Downloads/tenacity-8.4.0-py3-none-any.whl
tenacity/__init__.py
tenacity/_utils.py
tenacity/after.py
tenacity/before.py
tenacity/before_sleep.py
tenacity/nap.py
tenacity/py.typed
tenacity/retry.py
tenacity/stop.py
tenacity/tornadoweb.py
tenacity/wait.py
tenacity-8.4.0.dist-info/LICENSE
tenacity-8.4.0.dist-info/METADATA
tenacity-8.4.0.dist-info/WHEEL
tenacity-8.4.0.dist-info/top_level.txt
tenacity-8.4.0.dist-info/RECORD
```

[...]

PR #473, opened 2024-06-17
Fix import error

istrebitel-1:
- If library installed from pypi, import were not correct

### DIFF
--- a/tenacity/__init__.py
+++ b/tenacity/__init__.py
@@ -650,7 +650,7 @@ def retry(*dargs: t.Any, **dkw: t.Any) -> t.Any:
         return wrap
 
 
-from tenacity._asyncio import AsyncRetrying  # noqa:E402,I100
+from tenacity.asyncio import AsyncRetrying  # noqa:E402,I100
 
 if tornado:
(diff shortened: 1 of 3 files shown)


tenacity/__init__.py
```diff
@@ -650,7 +650,7 @@ def wrap(f: WrappedFn) -> WrappedFn:
         return wrap
 
 
-from tenacity.asyncio import AsyncRetrying  # noqa:E402,I100
+from tenacity._asyncio import AsyncRetrying  # noqa:E402,I100
```
maximlt, line 653: This is wrong, there's no `_asyncio` module in the repo.

istrebitel-1: so, you are right, but I still unable to import from installed lib
image
image

istrebitel-1: it's in library generation, asyncio module now not creating

istrebitel-1:
> I have narrowed it down to be due to the setuptools configuration: [#471 (comment)](https://github.com/jd/tenacity/issues/471#issuecomment-2172738891)

yes, got same solution

jd:
Should be fixed by #474

PR #479, opened 2024-06-24
fix: Avoid overwriting local contexts with retry decorator

### LINKED ISSUE #478: Retry in recursive context is broken since 8.3.0
Hello
We observed breaking change in `@retry` decorator behavior after upgrading the library from 8.2.3 to 8.3.0/8.4.1 version.

Assuming the next code sample (simplified and adjusted to illustrate the exact use-case):
```python
from tenacity import RetryCallState, retry

MAX_RETRY_FIX_ATTEMPTS = 2


def do_retry(retry_state: RetryCallState):
    ex = retry_state.outcome.exception()
    _subject_: str = retry_state.args[0]

    if _subject_ == 'Fix':  # no retry on fix failure
        return False

    if retry_state.attempt_number >= MAX_RETRY_FIX_ATTEMPTS:
        return False

    if ex:
        do_fix_work()
        return True

    return False


@retry(reraise=True, retry=do_retry)
def _do_work(subject: str):
    if subject == 'Error':
        print(f'{subject} is not working')
        raise Exception(f'{subject} is not working')
    print(f'{subject} is working')


def do_any_work(subject: str):
    _do_work(subject)


def do_fix_work():
    _do_work('Fix')


if __name__ == '__main__':
    do_any_work('Error')
```

In version 8.2.3 it worked correctly means `do_any_work` function was called twice with `do_fix_work` in the middle and produced the next output:
```bash
Error is not working
Fix is working
Error is not working
...call-stack...
Exception: Error is not working
```

After upgrading to 8.3.0 or 8.4.1 the code is not working the same way.
There's no additional call of `do_any_work` after `do_fix_work` and the code does not end with error:
```bash
Error is not working
Fix is working

Process finished with exit code 0
```

hasier:
Fixes #478

After #434 we rely on a local context variable to keep track of the remaining actions in the stack. However, if a decorated function is called recursively within the retry context, the same variable is used and it is overwritten, therefore preventing the calls higher up the stack of completing successfully. With these changes we now copy the retry context each time the function is called so that we have a fresh new instance to work on.

### DIFF
--- a/releasenotes/notes/fix-local-context-overwrite-94190ba06a481631.yaml
+++ b/releasenotes/notes/fix-local-context-overwrite-94190ba06a481631.yaml
@@ -1,0 +1,4 @@
+---
+fixes:
+  - |
+    Avoid overwriting local contexts when applying the retry decorator.
--- a/tenacity/__init__.py
+++ b/tenacity/__init__.py
@@ -329,13 +329,19 @@ class BaseRetrying(ABC):
             f, functools.WRAPPER_ASSIGNMENTS + ("__defaults__", "__kwdefaults__")
         )
         def wrapped_f(*args: t.Any, **kw: t.Any) -> t.Any:
-            return self(f, *args, **kw)
+            # Always create a copy to prevent overwriting the local contexts when
+            # calling the same wrapped functions multiple times in the same stack
+            copy = self.copy()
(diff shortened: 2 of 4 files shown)


MattGal:
Having fixed code that was broken due to assumptions on the location of `statistics` today, FYI you might note this set of changes was breaking somewhere.

Issue #432, opened 2024-01-24
Format seconds in `before_sleep_log` similar to `after_log`


aguinane:
The `before_sleep_log` messages shows the amount of seconds until retry to 15 decimal places. 

The [`after_log`](https://github.com/jd/tenacity/blob/99e7482e8b027f0521315e0f7830a0badae5c561/tenacity/after.py#L31) function rounds the number of seconds to 3 decimals. 
The [`before_sleep_log`](https://github.com/jd/tenacity/blob/99e7482e8b027f0521315e0f7830a0badae5c561/tenacity/before_sleep.py#L31) function should do the same.

jd:
Already fixed — `before_sleep_log` now accepts a `sec_format` parameter (defaults to `"%.3g"`), matching `after_log`.

PR #489, opened 2024-08-09
Apply formatting to num seconds in before_sleep_log

aguinane:
Resolves issue [#432](https://github.com/jd/tenacity/issues/432)


tenacity/before_sleep.py
```diff
@@ -32,6 +32,7 @@ def before_sleep_log(
     logger: "logging.Logger",
     log_level: int,
     exc_info: bool = False,
+    sec_format: str = "%0.3f",
```
jd, line 35:
```suggestion
    sec_format: str = "%0.gf",
```
WDYT?

aguinane: Yep, that's better as it doesn't require changing the test cases either. I will repush.

aguinane:
Updated with %.3g instead or %0.3f as the default.

asqui:
This latest change deviates from the default format for `after_log()`:
https://github.com/jd/tenacity/blob/e2482a6f90c432992d2f094774131531d5983b75/tenacity/after.py#L31-L35

Should we update that too, so that the two are consistent?

jd:
Sounds like a good idea.

aguinane:
Made the change to `after_log` also.

jd (approved):

aguinane:
I've re-synced this with main. Anything else that needs to be updated?

Issue #495, opened 2024-09-09
Add Dictionary Configuration Support


kevdog824:
PREFACE: I searched open issues for a similar issue but did not find one. Feel free to mark as duplicate if this issue already exists.

I am proposing adding dictionary-based or static configuration-based functionality to Tenacity. I am not sure that this is the best way to describe the functionality I am hoping to see so let me elaborate below.

A convenient part of the standard library logging module is the `dictconfig` function that allows you to setup configuration for logging from a dictionary. This enables a wide range of options for configuring logging. For example:
- Statically defining the logging configuration via pyproject.toml or some other config file
- Allowing a framework or application runner to easily accept a custom logging definition without custom code ([see uvicorn's command line switches](https://www.uvicorn.org/settings/#logging) for an example of this)

I propose adding a functionality similar to `dictconfig` to tenacity for creating and configuring Retry objects. I think this would add some good benefits and development options for developers using tenacity.

jd:
Tenacity's decorator/constructor API is already quite flexible. A dictconfig-style approach would add significant complexity for a niche use case that can be achieved with a small wrapper in user code.

PR #498, opened 2024-09-28
Adding DictConfig

SalvatoreZagaria:
# Pull Request Description

This pull request includes new features for configuring retry parameters using the `DictConfig` class.

## Summary of Changes:

1. **Add Unit Tests for Tenacity's Default Configurations**:
    - Introduced a new test class `TestRetryDefaults` to validate the functionalities of `dict_config` within tenacity.
    - Created unit tests that cover:
        - Setting, getting, overriding, and deleting configuration attributes.
        - Testing retry behavior with both default and overridden configurations.

2. **Refactor Configuration Handling**:
    - Removed redundant attribute checks in the `__init__` method and simplified its initialization.
    - Added a new `get` method to access configuration values more explicitly and efficiently.

3. **Add Configurable Retry Parameters with `DictConfig`**:
    - Introduced a `DictConfig` class to manage default retry parameters in a singleton dictionary.
    - Enhanced flexibility by allowing default configurations to be overridden or extended when calling the retry decorator.

This is addressing issue #495

### DIFF
--- a/tenacity/__init__.py
+++ b/tenacity/__init__.py
@@ -26,6 +26,7 @@ from abc import ABC, abstractmethod
 from concurrent import futures
 
 from . import _utils
+from .config import dict_config
 
 # Import all built-in retry strategies for easier usage.
 from .retry import retry_base  # noqa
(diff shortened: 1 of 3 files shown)


jd (commented):
I don't think this is an infrastructure we want to offer in tenacity.

tk512:
> I don't think this is an infrastructure we want to offer in tenacity.

Is there another way / a compromise for this to be supported by Tenacity? I think it would be highly useful to have safer retry defaults, for one. Perhaps configurable via `pyproject.toml` as well.

PR #512, opened 2025-01-31
Ensure `before_sleep_log` message isn't multiline

sebastian-correa:
Some exceptions add extra `\n` characters to their message, causing the last . in the log to be a new line.


jd:
I'm failing to see how this is a problem in tenacity itself and not in the so called exception though

sebastian-correa:
@jd the PR's title might be a bit overreaching, since it won't solve actual multiline string representations (i.e., if the exception has `\n` "in between"). However, since the `tenacity` log adds a. `` at the end, I think the expectation (when implemented) was that it would come immediately after the exception string, which might not be the case if it ends with `\n.`

I do agree that the actual problem is more probably in the exception, but `tenacity` logs end up split across multiple lines. I thought about implementing my own `before_sleep_log`, but since it uses a private function, I desisted.

jd:
Right; I'd rather change the log line to not have a final `.` but using something like `{verb}: {e}`

PR #518, opened 2025-03-03
fix: return "Self" from "BaseRetrying.copy"

ThirVondukr:
Currently `BaseRetrying.copy` returns `BaseRetrying`, making reusing a "template"/base retrying object harder.
Example usecase:
```py
from tenacity import AsyncRetrying


class SomeClient:
    def __init__(self, retry: AsyncRetrying) -> None:
        self._retry = retry

    async def method(self) -> None:
        async for attempt in self._retry.copy():
            with attempt:
                pass  # Do something
```
```sh
$ mypy main.py 
main.py:9: error: "BaseRetrying" has no attribute "__aiter__" (not async iterable)  [attr-defined]
Found 1 error in 1 file (checked 1 source file)
```

### DIFF
--- a/tenacity/__init__.py
+++ b/tenacity/__init__.py
@@ -88,6 +88,8 @@ except ImportError:
 if t.TYPE_CHECKING:
     import types
 
+    from typing_extensions import Self
+
     from . import asyncio as tasyncio
     from .retry import RetryBaseT
     from .stop import StopBaseT
@@ -255,7 +257,7 @@ class BaseRetrying(ABC):
         retry_error_callback: t.Union[
             t.Optional[t.Callable[["RetryCallState"], t.Any]], object
         ] = _unset,
-    ) -> "BaseRetrying":
+    ) -> "Self":
         """Copy this object with some parameters changed if needed."""
         return self.__class__(
             sleep=_first_set(sleep, self.sleep),


ThirVondukr:
@jd On an unrelated note - would a PR that adds a package manager (uv seems to be a good choice nowadays) to setup a dev environment be welcome?
Also would be nice to set up https://docs.pypi.org/trusted-publishers/

tenacity/__init__.py
```diff
@@ -88,6 +88,8 @@
 if t.TYPE_CHECKING:
     import types
 
+    from typing_extensions import Self
```
jd, line 91: This is not in tenacity dependency list; you'll need to add it.

jd:
> @jd On an unrelated note - would a PR that adds a package manager (uv seems to be a good choice nowadays) to setup a dev environment be welcome? 

Yes, if that can replace tox I'll be happy.

> Also would be nice to set up https://docs.pypi.org/trusted-publishers/

I'll work on that at some point, good thing.

tenacity/__init__.py
ThirVondukr: I'm honestly not sure if it should be in runtime dependencies, mypy seems to define `typing_extensions>=4.6.0` in it's own dependencies, but I'm not entirely sure about IDEs/Editors and Pyright

ThirVondukr:
> > @jd On an unrelated note - would a PR that adds a package manager (uv seems to be a good choice nowadays) to setup a dev environment be welcome?
> 
> Yes, if that can replace tox I'll be happy.

If you need to run multiple python environments/versions/dependency versions tox is probably the right choice, but that could instead be done in CI and mypy/ruff seem to work correctly when you specify target version, though you still can run them in CI with all python versions.

tenacity/__init__.py
jd: Ah I missed it's in only in `TYPE_CHECKING`. I guess I don't know either. We can try this way and see what it breaks? Anyhow we'll remove it when we drop Python < 3.11 🤷

jd (approved):

ThirVondukr:
I've checked with mypy, pyright and PyCharm:
PyCharm seems to derive copied type correctly:
image
Mypy works fine:
```py
from tenacity import AsyncRetrying, Retrying

retrying = AsyncRetrying()
reveal_type(retrying.copy()) # note: Revealed type is "tenacity.asyncio.AsyncRetrying"

retrying_sync = Retrying()
reveal_type(retrying_sync.copy()) # note: Revealed type is "tenacity.Retrying"
```
Pyright too:
```py
src\main.py:4:13 - information: Type of "retrying.copy()" is "AsyncRetrying"
src\main.py:7:13 - information: Type of "retrying_sync.copy()" is "Retrying"
```
Typing extensions has section in documentation about using types in runtime: https://typing-extensions.readthedocs.io/en/latest/#runtime-use-of-types
It would only be a problem here if BaseRetrying.copy would be inspected for some reason:
```py
import typing

from tenacity import AsyncRetrying, Retrying

typing.get_type_hints(AsyncRetrying.copy)
```
but is seems to fail earlier due to `StopBaseT` not being defined since it's imported in a type-checking block too, so adding `Self` without `typing_extensions` doesn't seem to be a problem here.

PR #552, opened 2026-02-05
chore: drop Python 3.9 support (EOL)

Zac-HD:
Remove Python 3.9 from CI matrix, setup.cfg classifiers, tox envlist, mergify checks, and bump python_requires to >=3.10.

Inline code that was gated behind sys.version_info checks:
- Use slots=True directly in dataclass decorator (3.10+)
- Use futures.Future[t.Any] directly (3.9+)
- Remove redundant version checks and imports in tests

(extracted from https://github.com/jd/tenacity/pull/551)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -18,8 +18,6 @@ jobs:
     strategy:
       matrix:
         include:
-          - python: "3.9"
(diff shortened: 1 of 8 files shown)


jd:
@mergify queue

PR #626, opened 2026-03-07
Feat: add retry_unless_exception_cause_type (#625)

### LINKED ISSUE #625: [FEATURE REQUEST] retry_unless_exception_cause_type
`retry_if_exception_cause_type` is an useful `retry_base` class but the inverted variant is missing.
It would be nice to have a `retry_base` class that retries if the exception is anything other than the specified exception(s).

An alternative is to implement an `__invert_` overload on `retry_base` but not sure how viable that is.

Kitan-Dara06:
This PR introduces the retry_unless_exception_cause_type class to mirror the existing retry_unless_exception_type logic. It halts the retry loop immediately if the root cause of the raised exception matches the specified type.

Related Issue:
Resolves #625

Testing:

[x] Implemented core logic in tenacity/retry.py

[x] Added unit tests in tests/test_tenacity.py confirming the retry loop halts on a matched cause and continues otherwise.

[x] Passed all local pytest checks (149/149).

### DIFF
--- a/tenacity/retry.py
+++ b/tenacity/retry.py
@@ -57,8 +57,26 @@ class retry_base(abc.ABC):
             return retry_any(*other.retries, self)
         return retry_any(other, self)
 
+    def __invert__(self) -> "retry_base":
+        """Return a retry strategy that is the logical inverse of this one."""
+        return _retry_inverted(self)
+
 
 RetryBaseT = retry_base | typing.Callable[["RetryCallState"], bool]
+
+
+class _retry_inverted(retry_base):
+    """Retry strategy that inverts the decision of another retry strategy."""
+
+    def __init__(self, retry: "retry_base") -> None:
+        self.retry = retry
+
+    def __call__(self, retry_state: "RetryCallState") -> bool:
+        return not self.retry(retry_state)
+
+    def __invert__(self) -> "retry_base":
+        # Double inversion returns the original.
+        return self.retry
 
 
 class _retry_never(retry_base):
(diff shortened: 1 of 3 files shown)


rgullik1:
The new class LGTM, but I don't really think the inverse overload behaves as one would expect to make it useful. 
Technically it does invert but it will retry when there's no exceptions. I don't really think that's useful. Would be better to just add  the new class IMO.

tenacity/retry.py
```diff
@@ -57,10 +57,28 @@ def __ror__(self, other: "RetryBaseT") -> "retry_any":
             return retry_any(*other.retries, self)
         return retry_any(other, self)
 
+    def __invert__(self) -> "retry_base":
+        """Return a retry strategy that is the logical inverse of this one."""
+        return _retry_inverted(self)
+
```
rgullik1, line 63:
```
```

tenacity/retry.py
```diff
@@ -57,10 +57,28 @@ def __ror__(self, other: "RetryBaseT") -> "retry_any":
...
+        self.retry = retry
+
+    def __call__(self, retry_state: "RetryCallState") -> bool:
+        return not self.retry(retry_state)
+
+    def __invert__(self) -> "retry_base":
+        # Double inversion returns the original.
+        return self.retry
+
+
```
rgullik1, line 81:
```
```

rgullik1 (changes requested):

tenacity/retry.py
rgullik1: I don't think the invert overload will be useful, please remove

tenacity/retry.py
rgullik1: Remove this also.

PR #629, opened 2026-03-11
feat: deprecate `initial` in favor of `multiplier` in `wait_exponential_jitter`

jd:
Every other exponential wait class uses `multiplier`. Having `initial`
only on `wait_exponential_jitter` is confusing. Add `multiplier` param
and deprecate `initial` as its alias.

- Add `multiplier` parameter (default 1)
- Emit `DeprecationWarning` when `initial` is used
- Raise `ValueError` if both `initial` and `multiplier` are specified

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/releasenotes/notes/deprecate-initial-for-multiplier-c7b4e2d9f1a83065.yaml
+++ b/releasenotes/notes/deprecate-initial-for-multiplier-c7b4e2d9f1a83065.yaml
@@ -1,0 +1,10 @@
+---
+deprecations:
+  - |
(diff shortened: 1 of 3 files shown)


jd:
This pull request is part of a stack:
1. feat: deprecate `initial` in favor of `multiplier` in `wait_exponential_jitter` ([#629](https://github.com/jd/tenacity/pull/629)) 👈

PR #633, opened 2026-03-14
Allow after hooks to override retry delay

muzakkirhussain011:
## Summary
- let after callbacks replace the next retry delay with a non-None return value
- normalize override values to float seconds and clamp negatives to zero
- resolve async override values before applying them in async-capable retry controllers

## Testing
- Not run locally in this environment because no usable Python interpreter was available from PowerShell.

### DIFF
--- a/tenacity/__init__.py
+++ b/tenacity/__init__.py
@@ -114,6 +114,7 @@ class IterState:
     retry_run_result: bool = False
     delay_since_first_attempt: int = 0
     stop_run_result: bool = False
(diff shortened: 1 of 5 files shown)


AhsanSheraz:
Nice to see server-hinted delays getting first-class treatment across the ecosystem — for
anyone comparing approaches, [pyresilience ](https://pypi.org/project/pyresilience/)which implements the same idea as a
`delay_func` hook (`retry_after_delay()` parses both Retry-After forms) since 0.4.0.

PR #635, opened 2026-03-14
Preserve typing for retry-wrapped callables

muzakkirhussain011:
## Summary
- preserve wrapped callable signatures across retry(), retry(fn), and wraps()
- type the retry and statistics helper attributes on wrapped callables
- keep retry_with returning a new wrapped callable with updated retry settings

## Testing
- Not run locally in this environment because no usable Python interpreter was available from PowerShell.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+test_6.sh text eol=lf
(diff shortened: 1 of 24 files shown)


jd (changes requested):
impossible to review and full of weird bash scripts

PR #649, opened 2026-07-03
retry operators: validate | and & operands on both sync and async retry_base

HrachShah:
## What

The `|` and `&` operators on `retry_base` and `async_retry_base`
accept either a retry strategy or a plain callable. A non-callable
non-retry_base value (e.g. `True | retry_always`, `retry & 42`,
`None & async_retry`) used to silently build a `retry_any` or
`retry_all` whose `__call__` later raised `TypeError: '<type>'
object is not callable` (sync) or `TypeError: object bool can't be
used in 'await' expression` (async) from inside the running retry
loop, far from the original misuse. The operators now raise a clear
`TypeError` naming the bad operand at the construction site, so
common typos like `retry or True` or `True | async_retry` fail fast
with an actionable message.

## Where

- `tenacity/retry.py`: `retry_base._validate_predicate` (static
  method), called from `__and__`, `__rand__`, `__or__`, `__ror__`.
- `tenacity/asyncio/retry.py`: same guard added to
  `async_retry_base.__and__`, `__rand__`, `__or__`, `__ror__`,
  and to the leaf branches of `retry_any.__ror__` and
  `retry_all.__rand__` (which are reached when Python's reflected
  operator dispatch goes through the subclass's override rather than
  the base class's).

## Tests

- 5 new sync tests in `tests/test_tenacity.py::TestRetryConditions::test_retry_composition_rejects_non_callable`
  covering `True | retry`, `retry | False`, `retry & 42`,
  `"nope" | retry`, and `None & retry`.
- 4 new async tests in `tests/test_asyncio.py::TestAsyncRetryComposition`:
  `test_async_retry_composition_rejects_non_callable` (mirrors the
  sync test with async operators), `test_async_retry_composition_accepts_valid_operands`
  (regression guard for valid compositions still building retry_any /
  retry_all), `test_async_retry_any_ror_validates_leaf_operand` and
  `test_async_retry_all_rand_validates_leaf_operand` (cover the
  subclass-operator leaf branches that don't go through
  `async_retry_base\.__ror__` / `__rand__`).
- `python3 -m pytest tests/` passes 171 tests (was 167).
- `ruff check tenacity/asyncio/retry.py tenacity/retry.py` clean.
- Verified manually: `True | tenacity.retry_always` raises
  `TypeError: Retry predicates must be retry_base instances or
  callables, got bool: True`; `True | tenacity.asyncio.retry.retry_if_exception(lambda exc: True)`
  raises the same; legitimate compositions
  `async_retry | sync_retry`, `async_retry | async_retry`,
  `async_retry | lambda` still build a `retry_any` with the
  expected operands.

## Release notes

`releasenotes/notes/reject-non-callable-retry-operands-7c4e9f2b1a3d8c5e.yaml`
(sync, added previously) and
`releasenotes/notes/fix-async-retry-composition-validates-operand-8f3a2b1e6c4d9a72.yaml`
(async, this commit) per the project's reno convention.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -34,7 +34,7 @@ jobs:
             task: mypy
     steps:
       - name: Checkout 🛎️
-        uses: actions/checkout@v7.0.0
+        uses: actions/checkout@v6.0.3
         with:
           fetch-depth: 0
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -12,7 +12,7 @@ jobs:
     permissions:
       id-token: write
     steps:
-      - uses: actions/checkout@v7.0.0
+      - uses: actions/checkout@v6.0.3
         with:
           fetch-depth: 0
           fetch-tags: true
--- a/releasenotes/notes/fix-async-retry-composition-validates-operand-8f3a2b1e6c4d9a72.yaml
+++ b/releasenotes/notes/fix-async-retry-composition-validates-operand-8f3a2b1e6c4d9a72.yaml
@@ -1,0 +1,17 @@
+---
+fixes:
+  - |
+    ``async_retry_base`` operators (``|``, ``&``, and their reflected
+    forms) now validate their operand at the operator site, matching
+    the behaviour already in place for the sync ``retry_base``. A
+    non-callable, non-``retry_base`` value used to silently construct
+    an ``async_retry_any``/``async_retry_all`` whose ``__call__``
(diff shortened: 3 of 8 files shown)


.github/workflows/ci.yaml
```diff
@@ -34,7 +34,7 @@ jobs:
             task: mypy
     steps:
       - name: Checkout 🛎️
-        uses: actions/checkout@v7.0.0
+        uses: actions/checkout@v6.0.3
```
jd, line 37: unrelated

jd (dismissed):

PR #657, opened 2026-07-16
Fix is_coroutine_callable for a partial over an async callable object

chuenchen309:
## Problem

`is_coroutine_callable` handles a `functools.partial` by pulling out `call.func` and running `iscoroutinefunction()` on it:

```python
partial_call = isinstance(call, functools.partial) and call.func
dunder_call = partial_call or getattr(call, "__call__", None)
return inspect.iscoroutinefunction(dunder_call)
```

When `call.func` is a callable **object** whose `__call__` is `async` (not a plain coroutine function), `iscoroutinefunction()` inspects the *instance* rather than its `__call__`, so the partial is reported as not-async:

```python
class AsyncClient:
    async def __call__(self, x): ...

is_coroutine_callable(AsyncClient())                     # True  (via the __call__ branch)
is_coroutine_callable(functools.partial(AsyncClient()))  # False ❌
```

The same object works when it isn't wrapped in a partial, so `partial(...)` silently breaks a case that otherwise works. The consequence is real: `retry(...)(functools.partial(async_callable_obj, ...))` is routed to the synchronous `Retrying` (`tenacity/__init__.py:752` picks the engine via `is_coroutine_callable`), calls the coroutine without awaiting it, and returns a **leaked, un-awaited coroutine** as the "result" (with a `RuntimeWarning: coroutine ... never awaited`).

## Fix

Recurse into the partial's target. This detects a partial over an async callable object, keeps the existing partial-over-coroutine-function and bare-object cases working, and additionally handles nested partials.

## Test plan

Extended `tests/test_utils.py::test_is_coroutine_callable` with `functools.partial(AsyncClass())` / `functools.partial(SyncClass())` (partial over the *instance*, which the existing `partial(AsyncClass().__call__)` case didn't cover).

- On `main` the new `partial_async_instance` assertion fails; with this change the full file passes.
- `pytest tests/` → 171 passed, 1 skipped.
- `ruff check` / `ruff format --check` / `mypy tenacity/_utils.py` all clean.
- Added a reno release note.

---
Authored with assistance from Claude Code (Anthropic); the bug, fix, and test were reviewed and verified locally by the contributor.

### DIFF
--- a/releasenotes/notes/fix-is-coroutine-callable-partial-async-object-3c1f9a2b7e4d6058.yaml
+++ b/releasenotes/notes/fix-is-coroutine-callable-partial-async-object-3c1f9a2b7e4d6058.yaml
@@ -1,0 +1,7 @@
+---
+fixes:
+  - |
+    Detect a ``functools.partial`` wrapping a callable object whose
+    ``__call__`` is asynchronous as a coroutine callable. Previously such a
+    partial was misrouted to the synchronous retry path, which returned an
+    un-awaited coroutine instead of retrying the awaited call.
--- a/tenacity/_utils.py
+++ b/tenacity/_utils.py
@@ -92,8 +92,12 @@ def is_coroutine_callable(call: typing.Callable[..., typing.Any]) -> bool:
         return False
     if inspect.iscoroutinefunction(call):
         return True
-    partial_call = isinstance(call, functools.partial) and call.func
-    dunder_call = partial_call or getattr(call, "__call__", None)  # noqa: B004
+    if isinstance(call, functools.partial):
+        # Recurse into the wrapped target so a partial over a callable *object*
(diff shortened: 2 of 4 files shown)


jd (dismissed):
This likely needs tests.

chuenchen309:
Thanks @jd! The PR already adds unit tests for exactly these cases in `tests/test_utils.py` — `partial_async_instance = functools.partial(AsyncClass())` and its sync counterpart, asserting `is_coroutine_callable` returns `True`/`False` respectively.

I've now also added a behavioral test in `tests/test_asyncio.py` that goes through the real retry path: without the fix, `await AsyncRetrying()(functools.partial(async_callable_obj))` takes the sync branch and stores the un-awaited coroutine as the result — the test asserts both the awaited return value and that the body actually ran, so it fails on `main` and passes with the fix. Pushed as f5b36e9.

PR #659, opened 2026-07-19
Fix retry_if_exception_cause_type infinite loop on cyclic exception cause chains

### LINKED ISSUE #658: retry_if_exception_cause_type spins forever when the exception cause chain contains a cycle (e.g. `raise e from e`)
When the exception cause chain contains a cycle, `retry_if_exception_cause_type` never returns — the predicate spins at 100% CPU and the retry can neither proceed nor stop.

The scan in `tenacity/retry.py` walks `__cause__` until it reaches `None`:

```python
exc = retry_state.outcome.exception()
while exc is not None:
    if isinstance(exc.__cause__, self.exception_cause_types):
        return True
    exc = exc.__cause__
```

A cyclic chain never reaches `None`. Because the spin happens inside the retry predicate, `stop_after_attempt` / `stop_after_delay` never get the chance to end the retry — stop conditions are only consulted after the predicate returns. The docstring's premise (“until finding an exception in the chain that has no `__cause__`”) doesn't hold for cyclic chains.

The simplest cycle is ordinary, legal Python:

```python
from tenacity import Retrying, retry_if_exception_cause_type, stop_after_attempt

def boom():
    try:
        raise ValueError("inner")
    except ValueError as e:
        raise e from e   # self-referential __cause__

r = Retrying(
    retry=retry_if_exception_cause_type(KeyError),
    stop=stop_after_attempt(2),
    reraise=True,
)
r(boom)   # never returns; 100% CPU
```

Two-node cycles (`a.__cause__ = b; b.__cause__ = a`, then `raise a`) hang the same way. For precedent: the stdlib `traceback` module explicitly guards against cause/context cycles when walking exception chains, so cyclic chains are treated as a real case in the ecosystem.

Reproduced on current `master` (c650fb4, v9.1.4); the loop is present in every release that ships this predicate.

Related note: the proposed `retry_unless_exception_cause_type` feature (#625 / #626) walks the cause chain the same way and would inherit this hang unless the walk is cycle-safe.

Fix is two lines (track seen exception ids; stop when the chain repeats). PR ready with a regression test covering both cycle shapes.

earfman:
Fixes #658.

## Problem
`retry_if_exception_cause_type.__call__` walks the `__cause__` chain until it reaches `None`. A cyclic chain — the simplest being ordinary `raise e from e` — never reaches `None`, so the predicate spins forever at 100% CPU. Because the spin is inside the retry predicate, `stop_after_attempt` / `stop_after_delay` never get a chance to end the retry: stop conditions are only consulted after the predicate returns. Two-node cycles (`a.__cause__ = b; b.__cause__ = a`) hang the same way. The stdlib `traceback` module guards against exactly these cause/context cycles when walking exception chains.

## Fix
Track the ids of exceptions already visited and stop the walk when the chain repeats:

```python
seen: set[int] = set()
exc = retry_state.outcome.exception()
while exc is not None and id(exc) not in seen:
    if isinstance(exc.__cause__, self.exception_cause_types):
        return True
    seen.add(id(exc))
    exc = exc.__cause__
```

Behavior on acyclic chains is unchanged; a cyclic chain now finishes scanning each distinct exception once and returns the correct result.

## Tests
Adds `test_retry_if_exception_cause_type_with_cause_cycle` covering both cycle shapes (`raise e from e` and a two-node cycle). The new test hangs indefinitely on current `main` without the fix (only bounded by an external timeout) and passes in milliseconds with it. Full suite: 172 passed, 1 skipped, 12 subtests.

## Notes
As flagged in #658, the proposed `retry_unless_exception_cause_type` feature (#625 / #626) walks the cause chain the same way — if this fix lands, that feature can reuse the cycle-safe walk.

Prepared with AI assistance and independently verified before submission (fresh clone, hang reproduced on HEAD c650fb4, fix reviewed, full suite green, regression test confirmed non-vacuous).

### DIFF
--- a/tenacity/retry.py
+++ b/tenacity/retry.py
@@ -176,10 +176,12 @@ class retry_if_exception_cause_type(retry_base):
             raise RuntimeError("__call__ called before outcome was set")
 
         if retry_state.outcome.failed:
+            seen: set[int] = set()
             exc = retry_state.outcome.exception()
-            while exc is not None:
+            while exc is not None and id(exc) not in seen:
                 if isinstance(exc.__cause__, self.exception_cause_types):
                     return True
+                seen.add(id(exc))
                 exc = exc.__cause__
 
         return False
--- a/tests/test_tenacity.py
+++ b/tests/test_tenacity.py
@@ -1441,6 +1441,31 @@ class TestDecoratorWrapper(unittest.TestCase):
         except NameError:
             pass
 
+    def test_retry_if_exception_cause_type_with_cause_cycle(self) -> None:
+        # A cyclic __cause__ chain (e.g. ``raise e from e``) must not hang the
+        # cause scan; the predicate should give up once the chain repeats.
+        def _raise_self_caused() -> None:
+            try:
+                raise ValueError("inner")
+            except ValueError as e:
+                raise e from e
+
+        def _raise_two_node_cycle() -> None:
+            first = KeyError("first")
+            second = OSError("second")
+            first.__cause__ = second
+            second.__cause__ = first
+            raise first
+
+        for raiser in (_raise_self_caused, _raise_two_node_cycle):
+            r = Retrying(
+                retry=tenacity.retry_if_exception_cause_type(NameError),
+                stop=tenacity.stop_after_attempt(2),
+                reraise=True,
+            )
+            with self.assertRaises((ValueError, KeyError)):
+                r(raiser)
+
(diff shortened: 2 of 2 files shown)


earfman:
Closing this — #658 was fixed by #660, which landed the same cycle-guard approach and is already merged. This is now a patch against a tree that has the fix, and it has picked up conflicts on top of that.

Nothing needed here. Thanks @MohammedAnasNathani for picking it up.

PR #663, opened 2026-07-26
feat: add retry_unless_exception_cause_type (cycle-safe cause walk)

### LINKED ISSUE #625: [FEATURE REQUEST] retry_unless_exception_cause_type

MohammedAnasNathani:
## Summary
Implements **#625** — `retry_unless_exception_cause_type`, the cause-chain mirror of `retry_unless_exception_type` / partner to `retry_if_exception_cause_type`.

Also hardens cause walking with a **cycle-safe** helper so `raise e from e` cannot hang the retry loop (**#658**).

## API
```python
@retry(retry=retry_unless_exception_cause_type(NameError), reraise=True)
def call():
    ...
```

- Keep retrying while no matching type appears in the `__cause__` chain
- Stop (do not retry) when a matching cause is found
- Successful outcomes follow the same convention as `retry_unless_exception_type`

## Tests
- Integration: stop immediately when cause is NameError
- Unit: predicate true/false for OSError-vs-NameError causes
- Cycle: self-referential `__cause__` terminates

Closes #625  
Related: #658

### DIFF
--- a/tenacity/__init__.py
+++ b/tenacity/__init__.py
@@ -54,6 +54,7 @@ from .retry import (
     retry_if_not_result,
     retry_if_result,
     retry_never,
+    retry_unless_exception_cause_type,
     retry_unless_exception_type,
 )
 
@@ -808,6 +809,7 @@ __all__ = [
     "retry_if_not_result",
     "retry_if_result",
     "retry_never",
+    "retry_unless_exception_cause_type",
     "retry_unless_exception_type",
     "sleep",
     "sleep_using_event",
--- a/tenacity/__init__.py
+++ b/tenacity/__init__.py
@@ -273,13 +273,13 @@ class BaseRetrying(ABC):
         retry: retry_base | object = _unset,
         before: t.Callable[["RetryCallState"], None] | object = _unset,
         after: t.Callable[["RetryCallState"], None] | object = _unset,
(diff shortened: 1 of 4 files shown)


MohammedAnasNathani:
Ready for review. Local: 170 tests pass (+ tornado skipped offline); CI green on 3.10–3.14 (check/lint/mypy). Implements #625 with cycle-safe cause walk (#658).

tenacity/__init__.py
```diff
@@ -272,13 +273,13 @@ def copy(
...
+        before_sleep: t.Callable[["RetryCallState"], None] | object | None = _unset,
         reraise: bool | object = _unset,
         retry_error_cls: type[RetryError] | object = _unset,
         retry_error_callback: t.Callable[["RetryCallState"], t.Any]
-        | None
-        | object = _unset,
-        name: str | None | object = _unset,
+        | object
+        | None = _unset,
+        name: str | object | None = _unset,
```
jd, line 282: This is just noise/churn AFAICT

jd (changes requested):

Issue #519, opened 2025-03-10
Statistics not set properly if retry fn is already wrapped


thelazydogsback:
For example:
```
@timed()
@retry(stop=stop_after_attempt(3))
def my_call(...):
  ...
```
After the call yields `my_call.statistics = {}` 

In this case the timed wrapper is before the target call, which I assume is receiving the statistics. (If we wanted to put the `timed` second and time each call inside the retry, this should work fine.)

There are of course may ways to work around this (an additional named, intermediate fn, etc.) but maybe there is a different way to code things so that the target of the retry always receives the statistics. (vs. one up the call stack?)

Also, as a possible option I suggest something like:
`@retry=(return_statistics_line)` which would return a tuple that would return the retry statistics and the final value, if any - this would not set anything on the fn object itself.
```
@retry(return_statistics_line=True, return_after_attempt(3))
def my_call(...): ...

stats, result = my_call(...)
```
This also feels nicer from a multi-threading point of view -- even though we have the GIL in standard python, it just feels cleaner to have this more functional approach as opposed to making sure you grabbed the statistics before you called any other I/O-bound operation that could cause the function to be re-entered on another thread/request and clobbering the statistics.

PR #668, opened 2026-08-04
Fix static typing of @retry decorated instance methods

### LINKED ISSUE #532: pyright/pylance typing error with tenacity "Unknown | Awaitable[Unknown]"
i get multiple warnings with typing is this a problem of the libery ?

```python
import logging
from abc import ABC, abstractmethod
from datetime import timedelta
from typing import Any

import aiohttp
from aiobreaker import CircuitBreaker
from aiohttp import ClientError, ClientTimeout
from tenacity import (
    retry,
    retry_if_exception_type,
    stop_after_attempt,
    wait_exponential,
)

from bot.const import (
    CIRCUIT_FAILURE_THRESHOLD,
    CIRCUIT_RECOVERY_TIMEOUT,
    HTTP_TIMEOUT,
    RETRY_MAX_ATTEMPTS,
)

logger = logging.getLogger(__name__)

# Circuit Breaker konfigurieren
breaker = CircuitBreaker(
    fail_max=CIRCUIT_FAILURE_THRESHOLD,
    timeout_duration=timedelta(seconds=CIRCUIT_RECOVERY_TIMEOUT),
)


class BaseApiClient(ABC):
    def __init__(self, session: aiohttp.ClientSession):
        timeout = ClientTimeout(total=HTTP_TIMEOUT)
        self._session = session
        self._session._default_timeout = timeout

    @retry(
        stop=stop_after_attempt(RETRY_MAX_ATTEMPTS),
        wait=wait_exponential(multiplier=1, min=1, max=10),
        retry=retry_if_exception_type(ClientError),
        reraise=True,
    )
    @breaker
    async def _get(self, url: str, **kwargs: Any) -> dict[Any, Any]:
        """
        Führt einen GET-Request mit:
          - Timeout aus HTTP_TIMEOUT
          - automatischen Retries bei Netzwerkfehlern
          - Circuit Breaker bei too many failures
        """
        try:
            async with self._session.get(url, **kwargs) as resp:
                resp.raise_for_status()
                return await resp.json()
        except Exception:
            logger.exception("API GET failed for URL: %s", url)
            raise

    @abstractmethod
    async def close(self) -> None:
        """Optionales Cleanup, falls nötig."""
        ...

```

```python
from typing import Any, Dict

from pydantic import BaseModel

from bot.const import WAIFU_API_BASE
from bot.helpers.integrations.base import BaseApiClient


class WaifuResponse(BaseModel):
    url: str


class WaifuApiClient(BaseApiClient):

    async def get_wave_gif(self) -> WaifuResponse:
        data: Dict[Any, Any] = await self._get(url=f"{WAIFU_API_BASE}/sfw/wave")
        return WaifuResponse.model_validate(data)

    async def close(self) -> None:
        # kein Cleanup nötig
        pass

```

```error
Der Typ von "data" ist teilweise unbekannt.
  Der Typ von "data" ist "Unknown | Awaitable[Unknown]".PylancereportUnknownVariableType
(variable) data: Dict[Any, Any]

Der Typ „Unknown | Awaitable[Unknown]“ kann dem deklarierten Typ „Dict[Any, Any]“ nicht zugewiesen werden
  Der Typ „Unknown | Awaitable[Unknown]“ kann dem Typ „Dict[Any, Any]“ nicht zugewiesen werden.
    „Awaitable[Unknown]“ kann „Dict[Any, Any]“ nicht zugewiesen werden.PylancereportAssignmentType


Kein Parameter mit dem Namen "url"PylancereportCallIssue
(function) url: Unknown
```
error translated to english:
```error
The type of "data" is partially unknown.
The type of "data" is "Unknown | Awaitable[Unknown]". PylancereportUnknownVariableType
(variable) data: Dict[Any, Any]

The type "Unknown | Awaitable[Unknown]" cannot be assigned to the declared type "Dict[Any, Any]".
The type "Unknown | Awaitable[Unknown]" cannot be assigned to the type "Dict[Any, Any]".
"Awaitable[Unknown]" cannot be assigned to "Dict[Any, Any]". PylancereportAssignmentType

No parameter named "url" PylancereportCallIssue
(function) url: Unknown
```

syzayd:
Fixes #532

## What changed and why

`_RetryDecorated`, the Protocol tenacity uses as the static return type of `@retry`-decorated callables, did not implement `__get__`. Without a descriptor, type checkers (mypy, pyright/pylance) treat `instance.method` the same as the unbound `Class.method`, so an ordinary call like `instance.method(value=1)` looks like it is missing `self`, and the return type resolves to `Unknown`/`Any` instead of the real return type. That matches the reporter's symptoms exactly: "No parameter named url" (keyword arguments no longer line up once `self` is expected positionally) and a return type that shows up as `Unknown | Awaitable[Unknown]`.

This reproduces with a plain synchronous method too, no third party decorator involved:

```python
class Foo:
    @retry(stop=stop_after_attempt(3))
    def double(self, value: int) -> int:
        return value * 2

f = Foo()
reveal_type(f.double)   # same type as Foo.double, self still required
```

This is purely a static analysis bug. At runtime the wrapper produced by `wraps()` is a real `functools.wraps`-decorated function, and real functions always bind `self` correctly, so nothing was broken for callers who are not running a type checker.

The fix adds `__get__` overloads to `_RetryDecorated`: accessed via the class it keeps returning the full unbound signature, accessed via an instance it returns `_RetryDecorated[..., R]`, i.e. self stripped, return type preserved. `Concatenate`-based self stripping is not expressible generically for an arbitrary already-captured `ParamSpec` with today's typing spec, so `...` is used for the bound parameter list, matching the pragmatic approach used elsewhere for generic method decorator stubs.

## Test plan

Added `tests/test_tenacity.py::TestRetryTyping::test_retry_decorated_method_keeps_bound_signature`, which decorates an instance method and calls it the normal bound way. Since this repo's `mypy --strict` gate also covers `tests/`, this test doubles as the regression check: I verified by temporarily reverting the `tenacity/__init__.py` change that mypy fails on this exact test with:

```
Missing positional argument "self" in call to "__call__" of "_RetryDecorated"  [call-arg]
```

With the fix restored:

```
$ mypy tenacity tests
Success: no issues found in 19 source files

$ pytest -q
174 passed, 1 warning, 12 subtests passed in 2.87s

$ ruff check .
All checks passed!

$ ruff format --check .
20 files already formatted
```

Also added a reno release note following the convention used for the comparable #519 typing fix.

## Note on the extra commit

The first CI run on this PR failed lint on 14 pre-existing `RUF036` violations across `tenacity/__init__.py`, `tenacity/asyncio/__init__.py`, and `tenacity/retry.py`, none of them on lines this PR touches. `ruff` is not version-pinned in `pyproject.toml`, and the last green run on `main` (same base commit as this branch) was on 2026-08-01, so a newer `ruff` release appears to have started enforcing `RUF036` ("None not at the end of a type union") in between. That means `main` would fail the same lint check if re-run today, independent of this PR.

Rather than leave the PR red over something unrelated, the second commit reorders `None` to the end of each flagged union (`ruff check --fix --unsafe-fixes .`, then `ruff format .`). Pure mechanical reordering, no behavior change - verified with a full `mypy` + `pytest` + `ruff` run after. Happy to drop that commit and rebase once `main` is fixed separately, if you'd rather keep it out of this PR.

### DIFF
--- a/releasenotes/notes/fix-decorated-method-binding-729ace6e58a9fefa.yaml
+++ b/releasenotes/notes/fix-decorated-method-binding-729ace6e58a9fefa.yaml
@@ -1,0 +1,11 @@
+---
+fixes:
+  - |
+    Fix static typing of ``@retry``-decorated instance methods. The
+    ``_RetryDecorated`` protocol did not implement ``__get__``, so type
+    checkers such as mypy and pyright/pylance treated ``instance.method`` the
+    same as the unbound ``Class.method``, requiring an explicit ``self``
+    argument and reporting an incorrect (``Unknown``/``Any``) return type for
+    ordinary calls like ``instance.method(value=1)``. This was purely a
+    static-analysis issue; the wrapper is a real function at runtime and
+    always bound correctly.
--- a/tenacity/__init__.py
+++ b/tenacity/__init__.py
@@ -670,6 +670,22 @@ class _RetryDecorated(t.Protocol[P, R]):
 
     def __call__(self, *args: P.args, **kwargs: P.kwargs) -> R: ...
 
+    # Without an explicit `__get__`, type checkers treat `_RetryDecorated` as a
+    # plain callable attribute rather than a descriptor, so accessing a
+    # decorated method through an instance keeps demanding the original
+    # unbound signature (including `self`). Declaring `__get__` here fixes
+    # attribute access on both the class and an instance; the runtime object
+    # is a real function (see `wraps` above), which already binds correctly,
+    # so this only affects static analysis. See issue #532.
+    @t.overload
+    def __get__(
+        self, instance: None, owner: type | None = None
+    ) -> "_RetryDecorated[P, R]": ...
+    @t.overload
+    def __get__(
+        self, instance: object, owner: type | None = None
+    ) -> "_RetryDecorated[..., R]": ...
+
 
 class _AsyncRetryDecorator(t.Protocol):
     @t.overload
--- a/tests/test_tenacity.py
+++ b/tests/test_tenacity.py
@@ -2196,6 +2196,28 @@ class TestRetryTyping(unittest.TestCase):
         self.assertEqual(with_raw_result, "1")
         self.assertEqual(with_constructor_result, "1")
 
+    def test_retry_decorated_method_keeps_bound_signature(self) -> None:
+        """A decorated instance method must type-check like a bound method.
+
+        Without a descriptor (``__get__``) on ``_RetryDecorated``, static type
+        checkers treat ``instance.method`` the same as the unbound
+        ``Class.method``, so a normal call with only the non-``self`` keyword
+        arguments looks like a type error and the return type resolves to
+        ``Any``/``Unknown``. This does not fail at runtime (functools.wraps
+        returns a real function, which Python always binds correctly), but it
+        does fail under `mypy --strict`, which also type-checks this file.
+        See issue #532.
+        """
+
+        class Doubler:
+            @retry(stop=tenacity.stop_after_attempt(3))
+            def double(self, value: int) -> int:
+                return value * 2
+
+        doubler = Doubler()
+        result: int = doubler.double(value=21)
+        self.assertEqual(result, 42)
+
 
 class TestMockingSleep:
     RETRY_ARGS = {


tenacity/asyncio/__init__.py
```diff
@@ -75,16 +75,16 @@ class AsyncRetrying(BaseRetrying):
     def __init__(
         self,
         sleep: t.Callable[
-            [int | float], None | t.Awaitable[None]
+            [int | float], t.Awaitable[None] | None
```
jd: looks like noise/churn here?

zaidwhy: Fair call, reverted. That commit was fixing an unrelated pre-existing RUF036 lint failure (ruff isn't pinned, a newer release started flagging 14 pre-existing type-union orderings repo-wide, unrelated to this PR's diff). Split it out into #669 so this PR stays focused on the #532 fix.

jd (dismissed):

zaidwhy:
Rebased onto main now that #670/#671 landed - picks up the ruff/mypy pins so CI is green again. This is ready to merge whenever you get a chance.

jd (approved):

PR #670, opened 2026-08-05
chore: pin ruff to 0.16.1

jd:
An unpinned ruff means every new ruff release can break CI with newly
stabilized rules. Pin it to the current version so upgrades are explicit
and deliberate.

Upgrading to 0.16.1 stabilizes RUF036 (None not at the end of a type
union); fix the 14 violations in the same commit via ruff's auto-fix so
the pinned version lints clean. The changes are pure annotation
reordering with no semantic effect.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>


jd:

| # | Pull Request | Link | |
|--:|---|---|---|
| 1 | chore: pin ruff to 0.16.1 | [#670](https://github.com/jd/tenacity/pull/670) | 👈 |
| 2 | chore: pin mypy to 2.3.0 | [#671](https://github.com/jd/tenacity/pull/671) |  |

PR #677, opened 2026-08-05
fix: annotate wait_base.__radd__ as taking the int seed from sum()

jd:
`__radd__` was annotated `other: wait_base`, which is the one type it can
never receive: `wait_base.__add__` always succeeds, so Python never falls
back to the reflected operator for two wait strategies. The only caller is
`sum()`, which seeds its accumulator with the int 0 — hence the
`type: ignore[comparison-overlap]` on `other == 0`, and the dead
`return self` branch behind it that `--warn-unreachable` flags.

Annotate the parameter as `int` and return `NotImplemented` for a non-zero
left operand, so `5 + wait_fixed(1)` raises `TypeError` at the addition
rather than building a `wait_combine` that explodes later when called.

With the signature corrected, mypy infers `sum([...])` over wait strategies
on its own, so four `type: ignore[list-item]` comments in
`test_wait_arbitrary_sum` are no longer needed.

Depends-On: #683


jd:

| # | Pull Request | Link | |
|--:|---|---|---|
| 1 | ci: run on pull requests targeting any base branch | [#683](https://github.com/jd/tenacity/pull/683) |  |
| 2 | fix: annotate wait_base.__radd__ as taking the int seed from sum() | [#677](https://github.com/jd/tenacity/pull/677) | 👈 |
| 3 | chore(mypy): enable the strictness options `strict` leaves off | [#678](https://github.com/jd/tenacity/pull/678) |  |
| 4 | refactor: drop always-true truthiness checks and enable truthy-bool | [#679](https://github.com/jd/tenacity/pull/679) |  |
| 5 | refactor: declare BaseAction's REPR_FIELDS and NAME as ClassVar | [#680](https://github.com/jd/tenacity/pull/680) |  |
| 6 | test: fix possibly-undefined and deprecated call sites | [#681](https://github.com/jd/tenacity/pull/681) |  |
| 7 | refactor: mark overridden methods with @override | [#682](https://github.com/jd/tenacity/pull/682) |  |

jd:
### Revision history

| 1 | initial | `b02c344` |  | 2026-08-05 15:39 UTC |
| 2 | rebase | `b02c344 → ae08acf` _(rebase only)_ |  | 2026-08-05 15:39 UTC |

PR #679, opened 2026-08-05
refactor: drop always-true truthiness checks and enable truthy-bool

jd:
`truthy-bool` reports objects tested for truthiness that implement neither
`__bool__` nor `__len__`, and so can only ever be true. Three sites:

`BaseRetrying._run_wait` and `AsyncRetrying._run_wait` both guarded the
wait call with `if self.wait:`. `wait` is typed `WaitBaseT` and defaults to
a `wait_none()` instance, so it is never falsy and the `sleep = 0.0` branch
has been dead since 17aefd9 -- a leftover from when the surrounding code
still used `if self.after is not None:` style guards. Call `self.wait`
unconditionally.

`if tornado:` guarded the optional import in two places. mypy only ever
sees the `try` branch, so it resolves the name to the module and reads the
test as always-true. Compute `_HAS_TORNADO` once and branch on that
instead; this also keeps `tornado.gen` fully typed, which annotating the
name as `ModuleType | None` would have thrown away.


jd:

| # | Pull Request | Link | |
|--:|---|---|---|
| 1 | refactor: drop always-true truthiness checks and enable truthy-bool | [#679](https://github.com/jd/tenacity/pull/679) | 👈 |
| 2 | refactor: declare BaseAction's REPR_FIELDS and NAME as ClassVar | [#680](https://github.com/jd/tenacity/pull/680) |  |
| 3 | test: fix possibly-undefined and deprecated call sites | [#681](https://github.com/jd/tenacity/pull/681) |  |
| 4 | refactor: mark overridden methods with @override | [#682](https://github.com/jd/tenacity/pull/682) |  |

jd:
### Revision history

| 1 | initial | `05deb5e` |  | 2026-08-05 15:39 UTC |
| 2 | rebase | `05deb5e → 7229e61` _(rebase only)_ |  | 2026-08-05 15:39 UTC |
| 3 | rebase | `7229e61 → 220001e` _(rebase only)_ |  | 2026-08-05 15:47 UTC |

PR #681, opened 2026-08-05
test: fix possibly-undefined and deprecated call sites

jd:
Two more error codes mypy leaves off under `strict`, both with real hits
in the test suite.

`possibly-undefined`: `test_retry_state` bound `retry_state` only inside
an `except ExtractCallState` block, then used it unconditionally. Had the
retry stopped raising, the test would have failed with `NameError` instead
of a useful assertion. Use `assertRaises` as a context manager, which both
asserts the exception is raised and binds the state unconditionally.

`deprecated`: `asyncio.iscoroutinefunction` is deprecated since 3.14 and
removed in 3.16, and was emitting a `DeprecationWarning` on every test run.
The line right below it already asserts the same property via
`inspect.iscoroutinefunction`, which is the documented replacement, so drop
the deprecated call rather than pin the suite to an API that is going away.


jd:

| # | Pull Request | Link | |
|--:|---|---|---|
| 1 | test: fix possibly-undefined and deprecated call sites | [#681](https://github.com/jd/tenacity/pull/681) | 👈 |
| 2 | refactor: mark overridden methods with @override | [#682](https://github.com/jd/tenacity/pull/682) |  |

jd:
### Revision history

| 1 | initial | `cdd9c7b` |  | 2026-08-05 15:39 UTC |
| 2 | rebase | `cdd9c7b → dbf3491` _(rebase only)_ |  | 2026-08-05 15:39 UTC |
| 3 | rebase | `dbf3491 → dc1546e` _(rebase only)_ |  | 2026-08-05 15:47 UTC |
| 4 | rebase | `dc1546e → d245e46` _(rebase only)_ |  | 2026-08-05 15:59 UTC |

PR #682, opened 2026-08-05
refactor: mark overridden methods with @override

jd:
`explicit-override` is off under `strict`, and 63 methods across the
library and tests were overriding a base method without saying so. That
makes a whole class of change silently lossy: rename or drop a method on
`retry_base`/`wait_base`/`stop_base`/`BaseRetrying` and every subclass keeps
its now-orphaned implementation, still importable, never called again.

`typing.override` only exists from Python 3.12, and tenacity has no runtime
dependencies, so `_utils` imports it from `typing_extensions` under
TYPE_CHECKING (as the module already does for `Self`) and falls back to a
small PEP 698 shim at runtime on 3.10/3.11. Type checkers only ever see the
`typing_extensions` name, so the check works regardless of the interpreter
mypy runs under.

That last point exposed a second problem worth fixing here: mypy's
`python_version` defaults to whatever interpreter runs it, so CI on 3.14 was
never verifying that the code is valid on the 3.10 we claim to support. Pin
`python_version = "3.10"` to match `requires-python`. It caught one real
case immediately -- `RetryCallState.__getstate__` is only an override from
3.11 on, where `object.__getstate__` was introduced, so it must not carry
the decorator.

Verified `mypy` reports the same result under both a 3.10 and a 3.14
interpreter, and the runtime shim sets `__override__` correctly on 3.10.

Depends-On: #681


jd:

| # | Pull Request | Link | |
|--:|---|---|---|
| 1 | test: fix possibly-undefined and deprecated call sites | [#681](https://github.com/jd/tenacity/pull/681) |  |
| 2 | refactor: mark overridden methods with @override | [#682](https://github.com/jd/tenacity/pull/682) | 👈 |

jd:
### Revision history

| 1 | initial | `0ef3a14` |  | 2026-08-05 15:39 UTC |
| 2 | rebase | `0ef3a14 → 73f02ad` _(rebase only)_ |  | 2026-08-05 15:39 UTC |
| 3 | rebase | `73f02ad → 8c5a928` _(rebase only)_ |  | 2026-08-05 15:47 UTC |
| 4 | rebase | `8c5a928 → c5fa908` _(rebase only)_ |  | 2026-08-05 15:59 UTC |

PR #695, opened 2026-08-05
fix: read the tornado global live instead of an import-time snapshot

jd:
#679 replaced `if tornado:` with a `_HAS_TORNADO` bool computed once at
import, to stop `truthy-bool` flagging the module object as always true.
But only one of the two call sites was converted: the `retry()` decorator
still dereferences the live `tornado` global right after testing the
snapshot. The two desync as soon as anything reassigns the global.

A/B verified against a2af454: `tenacity.tornado = None` followed by any
`@tenacity.retry` decoration returned normally before, and now raises
`AttributeError: 'NoneType' object has no attribute 'gen'`. Nulling the
module global is the standard way a downstream suite exercises the
non-tornado path without uninstalling tornado.

Test the availability of the live global at the point of use, and add a
regression test. `_HAS_TORNADO` is kept for the module-bottom import guard,
where snapshot and global cannot diverge.

Depends-On: #694


jd:

| # | Pull Request | Link | |
|--:|---|---|---|
| 1 | fix: restore support for falsy wait values and callable + strategy | [#694](https://github.com/jd/tenacity/pull/694) |  |
| 2 | fix: read the tornado global live instead of an import-time snapshot | [#695](https://github.com/jd/tenacity/pull/695) | 👈 |

tenacity/__init__.py
```diff
@@ -786,7 +786,10 @@ def wrap(f: t.Callable[P, R]) -> _RetryDecorated[P, R]:
         ):
             r = AsyncRetrying(*dargs, **dkw)
         elif (
-            _HAS_TORNADO
+            # Read the live global rather than the import-time `_HAS_TORNADO`
+            # snapshot: test suites routinely force the non-tornado path by
+            # setting `tenacity.tornado = None`, and the two must not desync.
+            tornado is not None  # type: ignore[redundant-expr]
```
jd, line 792: turn this into a function and drop the global?

jd: Done — `_HAS_TORNADO` is now a `_has_tornado()` function and the global is gone. Both call sites (the `retry()` dispatch and the module-bottom import guard) go through it, so they read the live `tornado` and can't drift apart, which was the actual bug here.

Bonus: it also drops the `# type: ignore[redundant-expr]` the constant needed. `redundant-expr` only fires on the left operand of an `and`, not on a plain `return`, so the function form needs no suppression at all.

Note the stack rebase after #694 merged replaced this PR — the change is now in #696.

PR #698, opened 2026-08-08
Add retry_base_exceptions to the negated exception-type predicates

### LINKED ISSUE #529: retry_if_not_exception_type swallows asyncio.CancelledError
# Problem

When using `retry_if_no_exception_type` to avoid retries for a specific exception, this causes `asyncio.CancelledError` to be caught and retried. This is contra-typical behaviour, as `CancelledError` does not typically cause a retry to occur in other scenarios. This is because task cancellation is not typically a retry case, and allow these errors to bubble up enables cancellation utilities like `asyncio.wait_for` to work as excpected.

# Steps to reproduce

Observe the behaviour as expected with this:

```
from tenacity import retry, wait_fixed, stop_after_attempt
import asyncio

@retry(
    wait=wait_fixed(10),
    stop=stop_after_attempt(2),
    reraise=True,
)
async def test():
    print("Sleeping")
    await asyncio.sleep(3)
    print("Done")

await asyncio.wait_for(test(), 1)  # Expected to cancel after 1s
 ```

The output here is:
```
Sleeping
CancelledError                            Traceback (most recent call last)
... (traceback)
```

We see incorrect behaviour in this case:

```
from tenacity import retry, wait_fixed, stop_after_attempt
import asyncio

@retry(
    wait=wait_fixed(10),
    stop=stop_after_attempt(2),
    reraise=True,
    retry=retry_if_not_exception_type(ValueError)
)
async def test_with_issue():
    print("Sleeping")
    await asyncio.sleep(3)
    print("Done")

await asyncio.wait_for(test_with_issue(), 1)  # Expected to cancel after 1s
 ```

The output here is:
```
Sleeping
Sleeping
Done
```

Here we see the retry occurring. We expect no retry to occur, and for the behaviour in this case to be the same as the former.

dchaudhari7177:
`retry_if_not_exception_type` and `retry_unless_exception_type` both decide with `not isinstance(e, self.exception_types)`, so anything outside the listed types is retried — including `asyncio.CancelledError`. Retrying a cancellation swallows it: under `asyncio.wait_for` the call is retried instead of unwinding, so `wait_for` never raises `TimeoutError`. `KeyboardInterrupt`, `SystemExit` and `GeneratorExit` behave the same way.

**This PR no longer changes that by default.** The first revision did, and @jd was right to reject it — as @Mukller's reproduction shows, it flips existing outcomes:

```
main: retry_if_not_exception_type(IOError)(CancelledError) -> True
```

So it is now opt-in, and the default is untouched:

```python
retry_if_not_exception_type(IOError, retry_base_exceptions=False)
retry_unless_exception_type(NameError, retry_base_exceptions=False)
```

`retry_base_exceptions` defaults to `True`, which is today's behaviour. With it left alone, every existing caller behaves exactly as before; the predicate does not even consult the new branch.

## What changed

- `retry.py`: a `retry_base_exceptions: bool = True` parameter on both predicates, and `_signals_control_flow` — the `not isinstance(e, Exception)` test — consulted only when it is `False`. Docstrings on both classes say what the flag does and that the default is the historical one.
- `tests/test_tenacity.py`: nine assertions across five tests.
  - Two pin that the **default still retries** `CancelledError`, `KeyboardInterrupt`, `SystemExit` and `GeneratorExit` on both predicates — i.e. that this is inert for existing callers.
  - Two pin that `retry_base_exceptions=False` makes all four propagate.
  - One runs both predicates under **both** settings against an ordinary `Exception`, pinning that the flag does not move the legs it must not touch.

`retry_if_exception_type` is unchanged and remains the way to retry a cancellation deliberately.

## Verification

Full suite on this branch: **174 passed, 12 subtests passed**. `ruff format --check` clean on both touched files; `ruff check` reports 14 `RUF036`, all pre-existing in lines this PR does not touch.

## Remaining question for the maintainer

The narrower thing I would still like to raise separately, and am happy to drop from here: `retry_if_exception_type` defaults to `Exception` and so already skips these types, while its two negated siblings do not. Which of the three predicates a user picks therefore decides whether `asyncio.wait_for` can time out. This PR gives them a supported way out without deciding that for them; whether the default should ever change is your call, and I am not asking for it here.

Closes #529


jd (dismissed):
This completely changes the contract of the API.

dchaudhari7177:
Fair — it does, and I should have proposed it rather than assumed it.

To be concrete about the alternative: the same fix can be opt-in and leave the current contract untouched, e.g.

```python
retry_if_not_exception_type(Boom, retry_base_exceptions=False)
```

defaulting to `True` (today's behaviour), so nothing changes for existing users and anyone hitting the cancellation case has a supported way out. That's a smaller change than what's here and doesn't touch anyone's semantics.

The reason I'd still like *something* to land: as it stands `retry_if_not_exception_type` turns an `asyncio.wait_for` timeout into a normal return — the caller never sees `TimeoutError` and can't tell the timeout happened. That one seems worth addressing on its own, separately from the `KeyboardInterrupt`/`SystemExit` question, if the broader change is off the table.

Happy to go any of three ways — rework as the opt-in flag above, narrow it to just the `wait_for`/cancellation case, or close this if you'd rather #529 be solved differently. Your call; I won't push further either way.

Mukller (commented):
Ran this branch locally (CPython 3.13 / Windows): the five new tests pass, and the full suite stays green at 133 passed + 12 subtests.

I also confirmed the behavior change against current main with the same `RetryCallState` fixture the tests use:

```
main:   retry_if_not_exception_type(IOError)(CancelledError) -> True
branch: False
main:   retry_unless_exception_type(NameError)(CancelledError) -> True
branch: False
```

So this does flip existing outcomes for anyone whose retries currently pass through `BaseException`s. I think the change is right — swallowing a `CancelledError` breaks `asyncio.wait_for` timeouts in ways that are painful to debug — but it deserves an explicit callout in the changelog as a behavior change rather than a pure fix, since `retry_unless_exception_type`'s documented contract ("retries unless X") has no carve-outs today.

Two smaller notes:

1. The helper's name/docstring say "signals control flow", but the implementation is exactly `not isinstance(e, Exception)` — any user-defined `BaseException` subclass that reports a failure (some frameworks use `BaseException` precisely to dodge blanket handlers) will now also be treated as control flow and not retried. Probably acceptable, but the docstring could state the actual rule.
2. This overlaps with #661 (`never retry CancelledError under retry_if_not_exception_type`) — worth coordinating so whichever shape lands first doesn't strand the other branch.

None of this blocks from my side; the tests cover all four classic `BaseException`s plus the explicit `retry_if_exception_type(CancelledError)` opt-in path.

dchaudhari7177:
You're right, and I should have said so in the description rather than filing it as a fix. This is a behaviour change to `retry_if_not_exception_type` and `retry_unless_exception_type`, not a bug fix — @Mukller's reproduction is exactly the shape of it:

```
main:   retry_if_not_exception_type(IOError)(CancelledError) -> True
branch: False
```

Anyone relying on "retries unless X" covering `BaseException` today gets different behaviour. That's a real contract break and it's your call, not mine.

The argument for making it anyway, stated plainly so you can reject it on the merits: `retry_if_exception_type` already defaults to `Exception` and so skips `CancelledError`, while its two negated siblings do not. Which of the three predicates you pick therefore decides whether `asyncio.wait_for` can time out — and that is not a decision anyone is making deliberately. But "it is surprising" is a much weaker claim than "it is wrong", and it does not by itself justify changing what a documented predicate means.

Three ways I can take this, in descending order of intrusiveness — happy with any, including the last:

1. **Keep the behaviour, document it as a break.** Changelog entry under a behaviour-change heading rather than a fix, and a docs note on both predicates. This is what @Mukller suggested and it's the minimum this needs if it lands at all.
2. **Make it opt-in.** `retry_if_not_exception_type(IOError, retry_base_exceptions=False)` defaulting to today's behaviour, so nothing changes for existing users and the asyncio case has a spelling. Costs a keyword argument on two classes; breaks nothing.
3. **Close it**, and leave #661 to solve the `CancelledError` case in whatever shape you prefer. This overlaps #661 anyway, and one of the two branches should be stranded deliberately rather than by accident.

Also worth fixing regardless of outcome: @Mukller is right that the helper's name oversells it. `_signals_control_flow` implies intent, but the implementation is `not isinstance(e, Exception)`, which also catches a user-defined `BaseException` subclass raised precisely to dodge blanket handlers. The docstring should state the actual rule.

Tell me which of the three and I'll push it, or close it yourself and I won't take it personally — the compatibility concern is legitimate.

dchaudhari7177:
@jd — reworked rather than re-argued. Pushed `63bc04f`, and retitled the PR, since the old title no longer describes it.

The default is now untouched. The skip is opt-in:

```python
retry_if_not_exception_type(IOError, retry_base_exceptions=False)
```

`retry_base_exceptions` defaults to `True` on both predicates, so every existing caller — including @Mukller's `retry_if_not_exception_type(IOError)(CancelledError) -> True` — behaves exactly as it does on main. I left the first commit in place so the diff between the two shows precisely what your objection removed.

Four of the nine new assertions exist only to pin that this is inert: they assert the **default still retries** `CancelledError`, `KeyboardInterrupt`, `SystemExit` and `GeneratorExit`. Four more pin the opt-in path, and one runs both predicates under both settings against an ordinary `Exception`, so the flag cannot quietly move a leg it should not.

174 passed, 12 subtests. `ruff format --check` clean on both files.

If you would rather not carry the parameter at all, say so and I will close this — the API surface is yours, and a stalled PR is worse than a closed one.

PR #700, opened 2026-08-23
fix: don't KeyError on statistics when begin() ran on another thread (#507)

### LINKED ISSUE #507: Statistics can be incorrectly initialized inside Temporal contexts
Within [temporal](https://docs.temporal.io/develop/python), an asynchronous python workflow manager, statistics are not being correctly initialized. Given that statistics are on a local thread, my best guess is that a different thread is calling "next_action" than the one that initializes the AsyncRetrying object. 

Temporal has a custom AsyncEventLoop, and a function to suspend a job and then "replay" it up to a point, which further complicates its interactions here.

My proposed solution is to either make statistics a normal attribute (I assume this would be pretty delicate), or to ensure we handle uninitialized elements of "statistics" robustly (see PR below)

```
{
  "message": "Activity failed: online_ingest with error Error from online_ingest_feature_source_activity: ISC Task failed with error: 'idle_for'",
  "stackTrace": "  File "/app/venv/lib/python3.10/site-packages/temporalio/worker/_workflow_instance.py", line 1644, in _run_top_level_workflow_function
    await coro

  File "/app/venv/lib/python3.10/site-packages/temporalio/worker/_workflow_instance.py", line 769, in run_workflow
    result = await self._inbound.execute_workflow(input)

  File "/app/venv/lib/python3.10/site-packages/ictemporal/internal/workflow_error_reporting_interceptor.py", line 47, in execute_workflow
    raise e

  File "/app/venv/lib/python3.10/site-packages/ictemporal/internal/workflow_error_reporting_interceptor.py", line 34, in execute_workflow
    return await super().execute_workflow(input)

  File "/app/venv/lib/python3.10/site-packages/temporalio/worker/_interceptor.py", line 328, in execute_workflow
    return await self.next.execute_workflow(input)

  File "/app/venv/lib/python3.10/site-packages/temporalio/worker/_workflow_instance.py", line 1916, in execute_workflow
    return await input.run_fn(*args)

  File "/app/venv/lib/python3.10/site-packages/ml/features/workflows.py", line 1026, in run
    raise ApplicationError(
",
  "cause": {
    "message": "Error from online_ingest_feature_source_activity: ISC Task failed with error: 'idle_for'",
    "stackTrace": "  File "/app/venv/lib/python3.10/site-packages/ml/features/workflows.py", line 1015, in run
    raise ApplicationError(
",
    "cause": {
      "message": "ISC Task failed with error: 'idle_for'",
      "stackTrace": "  File "/app/venv/lib/python3.10/site-packages/ml/features/workflows.py", line 1005, in run
    await online_ingest_activity

  File "/app/venv/lib/python3.10/site-packages/ml/features/workflows.py", line 551, in online_ingest_feature_source
    raise ApplicationError(
",
      "cause": {
        "message": "'idle_for'",
        "stackTrace": "  File "/app/venv/lib/python3.10/site-packages/ml/features/workflows.py", line 509, in online_ingest_feature_source
    async for attempt in retry_context:

  File "/app/venv/lib/python3.10/site-packages/tenacity/asyncio/__init__.py", line 166, in __anext__
    do = await self.iter(retry_state=self._retry_state)

  File "/app/venv/lib/python3.10/site-packages/tenacity/asyncio/__init__.py", line 153, in iter
    result = await action(retry_state)

  File "/app/venv/lib/python3.10/site-packages/tenacity/_utils.py", line 99, in inner
    return call(*args, **kwargs)

  File "/app/venv/lib/python3.10/site-packages/tenacity/__init__.py", line 428, in next_action
    self.statistics["idle_for"] += sleep
",
        "applicationFailureInfo": {
          "type": "KeyError"
        }
      },
      "applicationFailureInfo": {}
    },
    "applicationFailureInfo": {}
  },
  "applicationFailureInfo": {}
}
```

Mukller:
Fixes #507

## Root cause

`statistics` is a lazily initialized **thread-local** dict (`self._local`). `begin()` populates the four keys (`start_time`, `attempt_number`, `idle_for`, `delay_since_first_attempt`) on whichever thread starts the call. Under Temporal, the workflow can be replayed so that `begin()` runs on one thread while the retry loop (`iter()` → `next_action()`) is re-executed on another. The second thread's `_local.statistics` starts as an empty dict, and:

```python
self.statistics["idle_for"] += sleep   # KeyError: 'idle_for'
```

This matches the reporter's traceback exactly and their suspicion ("a different thread is calling next_action than the one that initializes").

I first tried initializing the property with a full default key set, but that breaks the documented contract that statistics "will be empty when the controller has never been ran" (and an existing pickle test asserts `restored.statistics == {}`), so this PR keeps the empty-until-begin semantics and makes the write path defensive instead.

## Changes

- `next_action()` now reads with defaults before incrementing:
  ```python
  stats = self.statistics
  stats["idle_for"] = stats.get("idle_for", 0) + sleep
  stats["attempt_number"] = stats.get("attempt_number", 1) + 1
  ```
  A fresh per-thread view therefore starts counting from attempt 1 — the same values a normal run would produce at that point.
- Regression test `TestStatisticsKeys::test_iter_on_fresh_thread_after_begin_elsewhere`: calls `begin()` on the main thread, then drives the public `iter()` from a worker thread with a failed outcome.

## Testing

- New test fails on unmodified source with exactly the reported `KeyError: 'idle_for'`; passes on this branch.
- Full suite: `pytest tests/test_tenacity.py -q` → 142 passed, 15 subtests passed (CPython 3.13 / Windows).
- `ruff check` / `ruff format --check` clean; reno note added under `releasenotes/notes/`.

### DIFF
--- a/releasenotes/notes/statistics-thread-local-keys-3f8a1c2d94b7e601.yaml
+++ b/releasenotes/notes/statistics-thread-local-keys-3f8a1c2d94b7e601.yaml
@@ -1,8 +1,8 @@
 ---
 fixes:
   - |
-    Initialize the per-thread ``statistics`` dict with the full key set
-    instead of an empty dict. When ``begin()`` ran on a different thread than
-    the retry loop (for example under Temporal's replay worker), reading
-    ``statistics["idle_for"]`` in ``next_action`` raised ``KeyError``.
-    See issue #507.
+    Fixed a :exc:`KeyError` on ``statistics["idle_for"]`` when ``begin()``
+    ran on a different thread than the retry loop (for example under
+    Temporal's replay worker): ``next_action()`` now reads ``idle_for`` and
+    ``attempt_number`` with defaults instead of assuming they were already
+    initialized on the current thread. See issue #507.
--- a/releasenotes/notes/statistics-thread-local-keys-3f8a1c2d94b7e601.yaml
+++ b/releasenotes/notes/statistics-thread-local-keys-3f8a1c2d94b7e601.yaml
@@ -1,0 +1,8 @@
+---
+fixes:
+  - |
+    Initialize the per-thread ``statistics`` dict with the full key set
+    instead of an empty dict. When ``begin()`` ran on a different thread than
+    the retry loop (for example under Temporal's replay worker), reading
+    ``statistics["idle_for"]`` in ``next_action`` raised ``KeyError``.
+    See issue #507.
--- a/tenacity/__init__.py
+++ b/tenacity/__init__.py
@@ -487,8 +487,13 @@ class BaseRetrying(ABC):
             sleep = rs.upcoming_sleep
             rs.next_action = RetryAction(sleep)
             rs.idle_for += sleep
-            self.statistics["idle_for"] += sleep
-            self.statistics["attempt_number"] += 1
+            # `begin()` may have run on a different thread than this action
+            # (e.g. Temporal's replay worker re-executing the workflow), in
+            # which case this thread's lazily created statistics dict starts
+            # empty. Read with defaults instead of raising KeyError (#507).
+            stats = self.statistics
+            stats["idle_for"] = stats.get("idle_for", 0) + sleep
+            stats["attempt_number"] = stats.get("attempt_number", 1) + 1
 
         self._add_action_func(next_action)
 
--- a/tests/test_tenacity.py
+++ b/tests/test_tenacity.py
@@ -18,6 +18,7 @@ import datetime
 import logging
 import pickle
 import re
+import threading
 import time
 import typing
 import unittest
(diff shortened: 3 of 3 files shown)


Mukller (commented):
﻿Independently verified this on Windows/CPython 3.13:

- Reverted only the `__init__.py` hunk and ran the new test against unmodified source - it fails exactly as reported (KeyError path), passes with the fix.
- Full suite green here too: 142 passed + 15 subtests.
- Read-path audit: lines with direct `statistics[...]` reads are only the two fixed in `next_action`; the `delay_since_first_attempt` write further down is a plain dict assignment, which is safe on an empty per-thread dict. So the fix covers all read-modify-write sites.

One thing to fix before merge: the reno note says "Initialize the per-thread ``statistics`` dict with the full key set instead of an empty dict", but that is the approach this PR deliberately rejected (it would break the documented empty-until-begin contract and the existing pickle assertion). The actual change keeps the empty default and makes `next_action` read with defaults. As written, the release notes describe a behavior that does not ship.

Suggested replacement for the note body:

```yaml
fixes:
  - |
    Don't raise KeyError in ``next_action`` when ``begin()`` ran on a different
    thread than the retry loop (for example under Temporal's replay worker).
    Each thread's lazily created ``statistics`` view now starts counting from
    defaults instead of assuming the keys already exist. See issue #507.
```

Everything else looks good to me.

PR #708, opened 2026-09-07
fix: unshallow RTD checkout so reno can resolve changelog versions

### LINKED ISSUE #514: docs/Changelog: 9.0 release notes are missing
The changelog for tenacity latest at https://tenacity.readthedocs.io/en/latest/changelog.html shows only release notes up to 8.2.2 while the released version on PyPI and on GitHub is 9.0.0.

This seems wrong to me.

G-ChandraSekhar:
Fixes #514.

Root cause and reproduction steps are in my comment on #514. Short version: RTD's `git fetch --depth 50` doesn't give `reno` enough history to resolve version boundaries, causing every doc build to fail. This adds a `post_checkout` step to unshallow the clone before Sphinx runs. Verified locally — reproduced the failure and confirmed this fixes it.


G-ChandraSekhar:
Friendly ping — CI is green and this is a small docs-build fix. Happy to adjust if anything needs changing

